PR #3, opened 2024-03-28
Add support for table initializers to gate keep pruning

joamaki:
It's often important to avoid pruning data before the table has been fully populated by writers. Add support for this in the form of table initializers that can be registered before the application starts. Make the reconciler pruning conditional on the table being fully initialized.

The initializers are stored in the table entry, which means that the initializer count is coupled to the snapshot and thus a reader won't think the table is initialized before it gets a snapshot that indeed is initialized.

Example:
```
  // Adding an initializer (called before start)
  txn := db.WriteTxn(myTable)
  done := myTable.RegisterInitializer(txn)
  txn.Commit()

  rtxn1 := db.ReadTxn()
  myTable.Initialized(rtxn1) == false

  txn = db.WriteTxn(myTable)
  done(txn)
  txn.Commit()

  rtxn2 := db.ReadTxn()
  myTable.Initialized(rtxn1) == false
  myTable.Initialized(rtxn2) == true
```

### DIFF
--- a/types.go
+++ b/types.go
@@ -31,6 +31,10 @@ type Table[Obj any] interface {
 	// NumObjects returns the number of objects stored in the table.
 	NumObjects(ReadTxn) int
 
+	// Initialized returns true if the registered table initializers have
+	// completed.
+	Initialized(ReadTxn) bool
+
 	// Revision of the table. Constant for a read transaction, but
 	// increments in a write transaction on each Insert and Delete.
 	Revision(ReadTxn) Revision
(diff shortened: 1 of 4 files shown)


types.go
```diff
@@ -79,6 +83,12 @@ type RWTable[Obj any] interface {
 	// write transaction return the fresh uncommitted modifications if any.
 	Table[Obj]
 
+	// RegisterInitializer adds an initializers to the table. Returns
+	// a function to mark the initializer done. Once all initializers are
+	// done, Table[*].Initialized() will return true.
+	// This should only be used before the application has started.
```
pippolo84, line 89: What about adding an additional warning regarding the need to call the "done" function in a write txn?
Even if it is already stated for each function in `RWTable` interface, it might not be immediate for the returned function, since the user doesn't necessarily know the inner mechanics of the initializers registration.

pippolo84 (approved):
LGTM! :rocket:

PR #4, opened 2024-04-01
Remove dependency on k8s.io code

twpayne:
Firstly, this project is really interesting, thank you, and I believe that statedb will be extremely useful as the core data store for many applications outside Cilium and Kubernetes.

However, the current dependency graph is huge and includes k8s.io/api, k8s.io/apimachinery, and k8s.io/client-go. This makes the project significantly less appealing as few non-Kubernetes projects want to pull in large sections of the huge Kubernetes codebase as a dependency, even if most of it will be eventually pruned by the linker.

This is a draft PR, and probably should not be merged, to indicate the effort required to make this code more generic:

* The Kubernetes reflector is simply removed, as this is application-specific code and not relevant to statedb's core functionality. It can be moved to the application using statedb.
* The single type (a generic set) from k8s.io/apimachinery/pkg/util/sets is vendored, removing the dependency on k8s.io/apimachinery.

This is a draft PR designed to encourage discussion, not to be merged.


joamaki:
Thanks! Agree on removing the k8s specific code so let's go for that. I'll move it into cilium/cilium or into a statedb utils repo.

I would not vendor sets.Set since we can instead just use `map[T]bool` or `map[T]struct{}` internally where needed. Also `CollectSet` now returns the wrong type and it's used in `cilium/cilium` in places that expect the `sets.Set` type. I think we can just drop `CollectSet` for now.

Could you refactor the remaining `sets.Set` uses to just use `map[T]bool`? It's mostly just test+example and `NewTable` so should be straightforward.

twpayne:
> Could you refactor the remaining `sets.Set` uses to just use `map[T]bool`? It's mostly just test+example and `NewTable` so should be straightforward.

Done.

twpayne:
I couldn't resist adding an extra commit to use `map[T]struct{}` instead of `map[T]bool` for internal set types. As you know, this reduces memory usage at the cost of slightly different ergonomics. However, memory usage is unlikely to be a significant factor here (all sets are currently temporary, either within a function or within a test).

PR #5, opened 2024-04-01
Use github.com/alecthomas/assert/v2 for tests

twpayne:
Another draft PR...

github.com/alecthomas/assert/v2 is a modern testing assertion library that uses Go generics. It has a smaller API than github.com/stretchr/testify but -- more importantly -- better type safety. Specifically, stretchr/testify performs many silent type conversions that may conceal type errors, whereas alecthomas/assert uses Go generics to ensure type safety at compile time.

This draft PR switches to the core statedb tests to use alecthomas/assert so you can see what the changes look like. I would be happy to port the remaining tests if you are interested.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -14,8 +14,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
+	"github.com/alecthomas/assert/v2"
 	"go.uber.org/goleak"
 
 	"github.com/cilium/hive"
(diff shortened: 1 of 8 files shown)


db_test.go
```diff
@@ -214,31 +213,31 @@ func TestDB_DeleteTracker(t *testing.T) {
 		txn.Commit()
 	}
 
-	assert.EqualValues(t, table.Revision(db.ReadTxn()), expvarInt(metrics.RevisionVar.Get("test")), "Revision")
-	assert.EqualValues(t, 3, expvarInt(metrics.ObjectCountVar.Get("test")), "ObjectCount")
-	assert.EqualValues(t, 0, expvarInt(metrics.GraveyardObjectCountVar.Get("test")), "GraveyardObjectCount")
```
twpayne, line 219: These changes are examples of alecthomas/assert's type safety. Previously strechr/testify was comparing `int64`s and `uint64`s.

joamaki:
I would like to stick with testify for now to align with `cilium/cilium`'s choice of test assertion library. It'll avoid pulling in a new dependency into the project.

twpayne:
Ack, makes sense.

PR #7, opened 2024-04-04
Add support for prefix searching

joamaki:
Add the Prefix() method for doing a prefix search on a table:

```
  var tbl Table[Foo]
  var index Index[Foo, string]
  tbl.Insert(wtxn, Foo{"foo"})
  tbl.Insert(wtxn, Foo{"foobar"})
  wtxn.Commit()

  iter, watch := tbl.Prefix(db.ReadTxn(), index.Query("fo"))
  // returns "foo" and "foobar", the watch channel will be the channel
  // of the parent of "foo" since that's the most specific one.
```

### DIFF
--- a/db.go
+++ b/db.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"runtime"
 	"slices"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -251,6 +252,9 @@ func (h Handle) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
 			table.sortableMutex().AcquireDuration(),
 		)
 	}
+
+	// Sort the table names so they always appear ordered in metrics.
+	sort.Strings(tableNames)
 
 	db.metrics.WriteTxnTotalAcquisition(
 		h.name,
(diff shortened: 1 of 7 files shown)


joamaki:
cc @tommyp1ckles

db.go
```diff
@@ -252,6 +253,8 @@ func (h Handle) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
 		)
 	}
 
+	sort.Strings(tableNames)
```
bimmlerd, line 256: nit: quick comment as to why this is useful?

db_test.go
```diff
@@ -198,7 +202,59 @@ func TestDB_LowerBound_ByRevision(t *testing.T) {
...
+
+	{
+		txn := db.WriteTxn(table)
+		table.Insert(txn, testObject{ID: 99, Tags: []string{"abcd"}})
+		txn.Commit()
+	}
+
+	select {
+	case <-watch:
+	case <-time.After(time.Second):
```
bimmlerd, line 251: If the CI machine skips a second due to virt hiccup or something, this becomes 50/50 afaik, since select choses from available channels pseudorandomly. Could check that `watch` isn't also closed before fatalling, even though that is also racy.

bimmlerd (approved):

db_test.go
joamaki: I'll bump the timeouts.

PR #10, opened 2024-04-09
Reuse the timer for the retry wait channel

joamaki:
Wait() was creating a new timer every time it was called. In high throughput scenarios this will cause a lot of unnecessary timers to be created.

To avoid this, use a single timer and reuse it by checking Stop() return value and reusing it if it has not fired yet.

### DIFF
--- a/reconciler/retries.go
+++ b/reconciler/retries.go
@@ -33,6 +33,8 @@ func newRetries(minDuration, maxDuration time.Duration, objectToKey func(any) in
 		queue:       nil,
 		items:       make(map[string]*retryItem),
 		objectToKey: objectToKey,
+		waitTimer:   nil,
+		waitChan:    make(chan struct{}),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


reconciler/retries.go
```diff
@@ -98,17 +112,27 @@ func (rq *retries) Add(obj any) {
 	}
 	item.object = obj
 	item.numRetries += 1
-	item.retryAt = time.Now().Add(rq.backoff.Duration(item.numRetries))
+	duration := rq.backoff.Duration(item.numRetries)
+	item.retryAt = time.Now().Add(duration)
 	heap.Push(&rq.queue, item)
+
+	// New item is at the top of the queue, reset the timer.
```
bimmlerd, line 119: nit: I don't think it's guaranteed that a new item is at the top after `Push`, but it's probably more efficient to reschedule anyway than trying to figure that out.

bimmlerd (approved):
Stared this for a while and I think it's correct.

PR #11, opened 2024-04-12
Introduce Changes() API

joamaki:
DeleteTracker() is confusing to users. It is essentially just providing an iterator for updates and deletes, so let's call it that: Changes().

```
var tbl Table[*Foo]
iter, err := tbl.Changes(db.ReadTxn())
if err != nil { return }
defer iter.Close()

for {
  for ev, rev, ok := iter.Next(); ok; ev, rev, ok = iter.Next() {
    if ev.Deleted {
      // ev.Object deleted
    } else {
      // ev.Object created or updated
    }

  }
  select {
  case <-ctx.Done():
    return
  case <-iter.Watch():
  }
}
```

Functional difference to before is that this does not allow "reiterating" as one could do with DeleteTracker's IterateWithError that stops on error and allows picking up iteration again to retry. But that's not needed with Changes() as the caller can take care of retrying the failed event before iterating again.

### DIFF
--- a/part/part_test.go
+++ b/part/part_test.go
@@ -5,6 +5,7 @@ package part
 
 import (
 	"encoding/binary"
+	"fmt"
 	"math/rand"
 	"testing"
 	"time"
@@ -54,24 +55,28 @@ func Test_search(t *testing.T) {
 	}
 }
 
-func intKey(n uint64) []byte {
+func uint64Key(n uint64) []byte {
 	return binary.BigEndian.AppendUint64(nil, n)
+}
+
+func uint32Key(n uint32) []byte {
+	return binary.BigEndian.AppendUint32(nil, n)
 }
 
 func Test_simple_delete(t *testing.T) {
 	tree := New[uint64]()
 	txn := tree.Txn()
 
-	_, hadOld := txn.Insert(intKey(1), 1)
+	_, hadOld := txn.Insert(uint64Key(1), 1)
 	require.False(t, hadOld)
 
-	_, hadOld = txn.Insert(intKey(2), 2)
+	_, hadOld = txn.Insert(uint64Key(2), 2)
 	require.False(t, hadOld)
 
-	_, hadOld = txn.Delete(intKey(1))
+	_, hadOld = txn.Delete(uint64Key(1))
 	require.True(t, hadOld)
 
-	_, _, ok := txn.Get(intKey(1))
+	_, _, ok := txn.Get(uint64Key(1))
 	require.False(t, ok)
 }
 
@@ -79,10 +84,10 @@ func Test_delete(t *testing.T) {
 	tree := New[uint64]()
 
 	// Do multiple rounds with the same tree.
-	for round := 0; round < 10; round++ {
+	for round := 0; round < 100; round++ {
 		// Use a random amount of keys in random order to exercise different
 		// tree structures each time.
-		numKeys := 10 + rand.Intn(5000)
+		numKeys := 10 + rand.Intn(1000)
 		t.Logf("numKeys=%d", numKeys)
 
 		keys := []uint64{}
@@ -98,9 +103,9 @@ func Test_delete(t *testing.T) {
 
 		txn := tree.Txn()
 		for _, i := range keys {
-			_, hadOld = txn.Insert(intKey(i), i)
+			_, hadOld = txn.Insert(uint64Key(i), i)
 			assert.False(t, hadOld)
-			v, _, ok := txn.Get(intKey(i))
+			v, _, ok := txn.Get(uint64Key(i))
 			assert.True(t, ok)
 			assert.EqualValues(t, v, i)
 		}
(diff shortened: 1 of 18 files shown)


joamaki:
I'm keeping the same pattern of providing the `WriteTxn` to `EventIterator()`. This way it's very clear how the modification is pushed into the database. We could also have `EventIterator` grab a `WriteTxn`, but that is wrought with deadlock dangers and it's less clear to the caller from what snapshot onwards will the events appear from. So I think it's still the way to go, even though it is confusing to user to take a WriteTxn when you're reading.

Hmm though maybe it could take a `ReadTxn` instead and then figure out does it need to internally do a `WriteTxn` against the target table or not... I'll try that!

types.go
```diff
@@ -71,12 +71,33 @@ type Table[Obj any] interface {
...
-	// DeleteTracker creates a new delete tracker for the table.
-	//
-	// It starts tracking deletions performed against the table from the
-	// current revision. A WriteTxn against the target table is required to
-	// add the tracker to the table.
-	DeleteTracker(txn WriteTxn, trackerName string) (*DeleteTracker[Obj], error)
+	// EventIterator returns an iterator for events happening on the table.
+	// This uses the revision index to iterate over the objects in the order
+	// they have changed. If 'includeDeleted' is true, then this registers a tracker
+	// to hold onto deleted objects until they are observed.
```
bimmlerd, line 77: If you want to go ahead with the `ReadTxn` then this should at least mention that if `includeDeleted` then you need to pass it a `WriteTxn` instead.

bimmlerd (commented):
I think I'm missing the benefit of taking a `ReadTxn` instead of a `WriteTxn` - is it that "logically" an event iterator should be a read-only operation? 

To me, the trade-off to clarity of what's needed in the iface is not worth it, and I'd advocate (again :grin:) for just taking a WriteTxn.

rastislavs:
Thanks, I think this naming is much better for users that do not look that much into statedb details and just want to use it.

Regarding `ReadTxn` vs `WriteTxn`, I agree that it's fine to just require a `WriteTxn` if it's mentioned in the docs why it is needed.
Another option would have two iterators:

```
UpdateEventIterator(txn ReadTxn)

DeleteEventIterator(txn WriteTxn) 
```

but not sure if this would not bring more confusion..

types.go
joamaki: You don't need to pass it a WriteTxn.

joamaki:
> I think I'm missing the benefit of taking a `ReadTxn` instead of a `WriteTxn` - is it that "logically" an event iterator should be a read-only operation?
> 
> To me, the trade-off to clarity of what's needed in the iface is not worth it, and I'd advocate (again 😁) for just taking a WriteTxn.

The benefit is that it's way less confusing to users. You just want to observe changes to the table, so why would you need to take a WriteTxn?

joamaki:
Hmm, would `ChangeIterator(...) (Iterator[Change], error)` be better than `Event`?

joamaki:
> Thanks, I think this naming is much better for users that do not look that much into statedb details and just want to use it.
> 
> Regarding `ReadTxn` vs `WriteTxn`, I agree that it's fine to just require a `WriteTxn` if it's mentioned in the docs why it is needed.
> 
> Another option would have two iterators:
> 
> ```
> UpdateEventIterator(txn ReadTxn)
> 
> DeleteEventIterator(txn WriteTxn) 
> ```
> 
> but not sure if this would not bring more confusion..

`DeleteEventIterator` to me implies it's only about the deletions.

joamaki:
Hey, I ended up changing this up a bit:
- EventIterator() is now called Changes() and returns a ChangeIterator
- The "includeDeleted" parameter is gone. If someone really just wants updates then one can just do LowerBound search on the revision index directly.

bimmlerd (approved):

rastislavs (approved):
Thanks again!

part/part_test.go
```diff
@@ -491,6 +514,142 @@ func Test_lowerbound(t *testing.T) {
...
+		_, _, tree = tree.Insert(int32Key(n), n)
+		keys = append(keys, n)
+	}
+	tree.PrintTree()
+
+	iter = tree.LowerBound(int32Key(0x20000))
+	for i := 0; i < 50; i++ {
+		n := uint32(0x20000 + i)
+		_, v, ok := iter.Next()
+		require.True(t, ok)
```
bimmlerd, line 607: `next(true, n)`?

part/part_test.go
```diff
@@ -491,6 +514,142 @@ func Test_lowerbound(t *testing.T) {
...
+		require.Equal(t, n, v)
+	}
+	_, _, ok := iter.Next()
+	require.False(t, ok)
+
+	iter = tree.LowerBound([]byte{})
+	for i := range keys {
+		_, v, ok := iter.Next()
+		require.True(t, ok)
```
bimmlerd, line 616: `next(true, keys[i])`?

part/part_test.go
```diff
@@ -491,6 +514,142 @@ func Test_lowerbound(t *testing.T) {
...
+	next(true, 0x010000)
+	next(false, 0)
+
+	iter = tree.LowerBound([]byte{0, 1, 0})
+	next(true, 0x010000)
+	next(false, 0)
+
+	// Node256
+	fmt.Println("node256:")
+	for i := 0; i < 50; i++ { // add less than 256 for some holes in node256.children
```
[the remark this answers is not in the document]
joamaki: :+1:

PR #13, opened 2024-04-15
[DRAFT] Rename Get() to Select() and First() to Get()

joamaki:
Playing around with slightly nicer query method names. WDYT?

---
Rename Get() to Select()/SelectWatch and First() to Get(), and FirstWatch to GetWatch().

```
  iter := tbl.Select(txn, Index.Query("foo"))
  iter, watch := tbl.SelectWatch(txn, Index.Query("foo"))
  iter, watch := tbl.SelectWatch(txn, Index.Query("foo"), func(obj *Object) bool { return obj.IsValid })

  obj, rev, ok := tbl.Get(txn, Index.Query("foo"))
  obj, rev, watch, ok := tbl.GetWatch(txn, Index.Query("foo"))
```

And add support for specifying filters alongside query:
```
  func (obj *Object) isValid() bool { ... }

  iter := tbl.Select(txn, Index.Query("foo"))
  iter = statedb.Filter(iter, Object.isValid)

  // above can be expressed as:

  iter := tbl.Select(
    txn,
    Index.Query("foo"),
    Object.isValid,
  )
```

### DIFF
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -347,7 +347,7 @@ func BenchmarkDB_RandomLookup(b *testing.B) {
 	for j := 0; j < b.N; j++ {
 		txn := db.ReadTxn()
 		for _, q := range queries {
-			_, _, ok := table.First(txn, q)
+			_, _, ok := table.Get(txn, q)
 			if !ok {
 				b.Fatal("object not found")
 			}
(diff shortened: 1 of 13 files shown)


joamaki:
Some further ideas:

```
type Table[T any] interface {
  ...
  Select[T](ReadTxn, Query[T]) Iterator[T]
  SelectWatch[T](ReadTxn, Query[T]) Iterator[T]
}
type Query[T] interface { ... }

// Base queries
func Prefix[T, K any])(Query[T], prefix K) Query[T]
func LowerBound[T any, K any](Index[T,K], minimum K) Query[T]
func Range[T, K any](Index[T, K], min K, max K) Query[T]
func All[T any](Index[T]) Query[T]

// Filters/selectors to narrow down queries
func Min[T any](Query[T]) Query[T]
func Max[T any](Query[T]) Query[T]
func First[T any](Query[T]) Query[T]
func Last[T any](Query[T]) Query[T]
func Filter[T any])(Query[T], func(T) bool) Query[T]
```

PR #17, opened 2024-04-22
reconciler: Add ability to provide default metrics

joamaki:
Noticed that we need the ability to provide the default metrics implementation in cilium/cilium as most reconcilers would not want their own specific implementation. It would be unnecessary boilerplate to have to depend on metrics and pass it in the configuration, so add it as an optional to Params[T].


bimmlerd (approved):
yeah looks sane
could add a line about this in the `Config` struct in `types.go`; metrics is the one field not documented and it's not immediately clear that one doesn't have to provide it to get the default

PR #19, opened 2024-04-23
Fix Changes() watch channel

joamaki:
When the iterator was exhausted it was refreshed regardless of whether the previous watch channel had closed or not. This lead to busy looping as it was constantly calling LowerBound() and returning a closed watch channel.

Fix this by checking whether the previous watch channel has closed and if not keep returning it to wait for actual changes.


bimmlerd (commented):
LGTM

PR #20, opened 2024-04-25
reconciler: Move Table from Params to Config

joamaki:
With Hive cannot currently have to private providers for the same type, even when they're in different scopes. To allow for the situation where the RWTable is provided privately in one module, but still accessible via a method for reconciliation use, move the Table from Params to Config so it can be provided directly and not via Hive.

The exact setup where this problem occurred was:
```
  // module 1
  var Cell = cell.Module(...,
    cell.ProvidePrivate(NewRWTableFoo),
    cell.Provide(NewWriter),
  ),
  type Writer struct {
    table RWTable[Foo]
  }
  func NewWriter(t RWTable[Foo]) *Writer
  func (w *Writer) UnsafeRWTable() RWTable[Foo] { return w.table }

  // module 2
  var Cell = cell.Module(...,
    cell.ProvidePrivate((*m1.Writer).UnsafeRWTable),
    cell.Invoke(reconciler.Register[Foo]),
  )
```
The above lead to:
```
  level=fatal msg="Failed to apply cell"
  error="cannot provide function \"...UnsafeRWTable: this function introduces a cycle
  (long error about NewWriter depending on RWTable[Foo] provided by UnsafeRWTable)
```

Hive's (uber/dig's) cycle detection doesn't seem to care about the private providers, and thinks the private provide of "module 2" can be reached by "NewWriter" in module 1. So to fix this and add more flexibility, just make the table a config parameter.

### DIFF
--- a/reconciler/benchmark/main.go
+++ b/reconciler/benchmark/main.go
@@ -122,9 +122,9 @@ func main() {
 			"test",
 			"Test",
 
-			cell.Provide(func(db_ *statedb.DB) (statedb.RWTable[*testObject], error) {
+			cell.Invoke(func(db_ *statedb.DB) error {
 				db = db_
-				return testObjects, db.RegisterTable(testObjects)
+				return db.RegisterTable(testObjects)
 			}),
 			cell.Provide(
 				func() (*mockOps, reconciler.Operations[*testObject]) {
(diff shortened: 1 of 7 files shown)


reconciler/benchmark/main.go
```diff
@@ -122,9 +122,9 @@ func main() {
 			"test",
 			"Test",
```
bimmlerd, line 124: commit msg typo:

`With Hive cannot currently have to private providers for the same type,
even when they're in different scopes.`

should probably be `With Hive _you_ cannot` and then `have _two_ private`

bimmlerd (approved):
bit unfortunate that dig doesn't understand this, but the API is fine either way

reconciler/benchmark/main.go
joamaki: Thanks!

PR #21, opened 2024-04-25
index: Add missing integer indexers

joamaki:
Add the missing integer indexers and simplify all by using Append*.


bimmlerd (commented):
lgtm ;)

PR #22, opened 2024-04-25
part: Fix bug with short and long key with shared prefix

joamaki:
If we insert "foobar" and then "foo" we ended up in a case where "foo" was the common prefix and the rest of the key become empty leading to out-of-bounds access to the key slice.

Handle the case where the key becomes empty when splitting a node by just inserting it as the leaf of the split node.

Added a regression test and validated it catched the bug:
```
--- FAIL: Test_prefix_regression (0.00s)
panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

goroutine 32 [running]:
testing.tRunner.func1.2({0x5f3380, 0xc00001a288})
        /home/jussi/sdk/go1.22.0/src/testing/testing.go:1631 +0x24a
testing.tRunner.func1()
        /home/jussi/sdk/go1.22.0/src/testing/testing.go:1634 +0x377
panic({0x5f3380?, 0xc00001a288?})
        /home/jussi/sdk/go1.22.0/src/runtime/panic.go:770 +0x132
github.com/cilium/statedb/part.(*Txn[...]).insert(0x6725e0, ...
        /home/jussi/go/src/github.com/cilium/statedb/part/txn.go:228 +0xdaf
```

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -52,24 +52,26 @@ const (
 )
 
 type fuzzObj struct {
-	id    uint64
+	id    string
 	value uint64
 }
 
-func mkID() uint64 {
-	return 1 + uint64(rand.Int63n(numUniqueIDs))
+func mkID() string {
+	// We use a string hex presentation instead of the raw uint64 so we get
+	// a wide range of different length keys and different prefixes.
+	return fmt.Sprintf("%x", 1+uint64(rand.Int63n(numUniqueIDs)))
 }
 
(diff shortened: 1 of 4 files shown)


bimmlerd:
Hm, race found another bug - investigating.

joamaki:
The second bug was that for non-leaf nodes we didn't return the old value.

bimmlerd (approved):

PR #23, opened 2024-04-30
part: Fix leaf-node related issues

joamaki:
See commits.

### DIFF
--- a/part/txn.go
+++ b/part/txn.go
@@ -181,23 +181,31 @@ func (txn *Txn[T]) insert(root *header[T], key []byte, value T) (oldValue T, had
 			break
 		}
 
-		// Clone the parent.
-		this = txn.cloneNode(this)
-		*thisp = this
-
 		child, idx := this.findIndex(key[0])
 		if child == nil {
 			// We've found a free slot where to insert the key.
 			if this.size()+1 > this.cap() {
 				// Node too small, promote it to the next size.
+				if this.watch != nil {
+					txn.watches[this.watch] = struct{}{}
+				}
 				this = this.promote(!txn.opts.rootOnlyWatch || this == newRoot)
+				txn.mutated.put(this)
+			} else {
+				// Node is big enough, clone it so we can mutate it
+				this = txn.cloneNode(this)
 			}
 			this.insert(idx, newLeaf(txn.opts, key, fullKey, value).self())
 			*thisp = this
 			return
 		}
+
+		// Clone the parent so we can modify it
+		this = txn.cloneNode(this)
+		*thisp = this
+		// And recurse into the child
 		thisp = &this.children()[idx]
-		this = child
+		this = *thisp
 	}
 
 	// A node exists where we wanted to insert the key.
--- a/part/txn.go
+++ b/part/txn.go
@@ -183,6 +183,9 @@ func (txn *Txn[T]) insert(root *header[T], key []byte, value T) (oldValue T, had
 					leaf.value = value
 				} else {
 					// This is a non-leaf node, create/replace the existing leaf.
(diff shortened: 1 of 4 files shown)


part/txn.go
```diff
@@ -157,101 +157,166 @@ func (txn *Txn[T]) cloneNode(n *header[T]) *header[T] {
...
-				this.insert(idx, newLeaf(txn.opts, key, fullKey, value).self())
-				*thisp = this
-				return
-			}
+		// Prefix matched. Consume it and go further.
+		key = key[len(this.prefix):]
+		if len(key) == 0 {
+			// Our key matches this node.
+			break
+		}
```
bimmlerd, line 182: Could also do something like - not sure if clearer.
```
		key, matched = bytes.CutPrefix(key, this.prefix)
		if !matched || len(key) == 0 {
			break			
		}
```

part/txn.go
```diff
@@ -157,101 +157,166 @@ func (txn *Txn[T]) cloneNode(n *header[T]) *header[T] {
...
+		*thisp = this
 
-			this.prefix = this.prefix[len(newPrefix):]
-			key = key[len(newPrefix):]
+		child, idx := this.findIndex(key[0])
+		if child == nil {
+			// We've found a free slot where to insert the key.
+			if this.size()+1 > this.cap() {
+				// Node too small, promote it to the next size.
+				this = this.promote(!txn.opts.rootOnlyWatch || this == newRoot)
```
bimmlerd, line 193: I checked whether the `promote`d node inherits the watch channel of where it is promoted from, but `promote` seems to allocate a new one. Why?

part/txn.go
```diff
@@ -157,101 +157,166 @@ func (txn *Txn[T]) cloneNode(n *header[T]) *header[T] {
...
+	default:
+		// Partially matching prefix, non-leaf node.
+		common := commonPrefix(key, this.prefix)
+
+		this = txn.cloneNode(this)
+		*thisp = this
+		this.prefix = this.prefix[len(common):]
+		key = key[len(common):]
+
+		newLeaf := newLeaf(txn.opts, key, fullKey, value)
```
bimmlerd, line 290: not that it's a big issue, but why do we end up storing the full key in leaf nodes again? Once we search towards this leaf, can't we reconstruct the key on the way down?

part/txn.go
```diff
@@ -157,101 +157,166 @@ func (txn *Txn[T]) cloneNode(n *header[T]) *header[T] {
...
+		newNode := &node4[T]{
+			header: header[T]{prefix: common},
+		}
+		newNode.setKind(nodeKind4)
+
+		switch {
+		case len(key) == 0:
+			newLeaf.prefix = common
```
bimmlerd, line 298: Why does the newLeaf need this prefix set? `newNodes`'s prefix is `common` already, so won't this "duplicate" the common part?

part/txn.go
joamaki: If we would construct the key on the way down we'd need to allocate to build it. There is very little cost to storing the key in the leaf nodes as we're already using the same underlying array in the prefix fields.

part/txn.go
joamaki: So that we can move it around and can rely on the prefix being correct. I guess we could just have it as nil and special-case when a leaf node is moved, but that's probably less easy to follow.

joamaki: Specifically this line wouldn't work: https://github.com/cilium/statedb/pull/23/files/754d4c402b1bb3f9dee86e8f953d06660081641d#diff-243d2f909d6df11ff7564c096de7f3fc5c336450a15def5587d8121b2fa86c6eR227. We'd need to carry the prefix over from the node pointing to it.

PR #24, opened 2024-05-01
reconciler: Remove StatusKindDelete

joamaki:
To be able to use multiple reconcilers against a single object we cannot do soft-deletes and have the reconciler delete the object, as then we would have multiple reconcilers wanting to delete the same object.

To fix this, drop the StatusKindDelete and use the ChangeIterator to iterate over both inserts and deletes.

The downside is that we can't observe failed deletions via the table, but it will be visible via module health. On the other hand, it is less confusing to users to not have deleted objects visible in the tables.

### DIFF
--- a/reconciler/reconciler.go
+++ b/reconciler/reconciler.go
@@ -5,7 +5,6 @@ package reconciler
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log/slog"
 	"time"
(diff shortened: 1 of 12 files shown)


reconciler/reconciler.go
```diff
@@ -1,11 +1,10 @@
-// SPDX-License-Identifier: Apache-2.0
+// SPDX-License-Identifier: Apache-2.0:
```
dylandreimerink, line 1: What is this?

dylandreimerink (approved):
Looks good to me

reconciler/reconciler.go
joamaki: oops. fixed.

txn.go
```diff
@@ -404,7 +406,7 @@ func (txn *txn) Commit() {
 	// If db is nil, this transaction has already been committed or aborted, and
 	// thus there is nothing to do.
 	if txn.db == nil {
```
bimmlerd, line 408: this check doesn't work any more, does it? Commit no longer zero's out the db.

PR #25, opened 2024-05-02
Set and Map and Docs!

joamaki:


twpayne (approved):
This is great, thank you! It makes StateDB much more accessible.

From a high level, the StateDB usage documentation is a long wall of text which needs to be read in sequence. Breaking it up into a few sections, e.g. initialization, reading, writing, and updating, would make it easier to navigate.

But thank you for writing this!

README.md
```diff
@@ -1,9 +1,509 @@
...
+wtxn.Commit()
+```
+
+## Persistent Map and Set
+
+The `part` package contains persistent `Map[K, V]` and `Set[T]` data structures.
+These, like StateDB, are implemented with the Persistent Adaptive Radix Trees.
+They are meant to be used as replacements for the built-in mutable Go hashmap
+in StateDB objects as they're persistent (operations return a copy) and thus 
+suitable to use in immutable objects.
```
bimmlerd, line 409: They're also more efficient to copy, compared to the built-in.

README.md
```diff
@@ -1,9 +1,509 @@
...
+pattern or a constructor function. 
+
+Also prefer persistent/immutable data structures within the object to avoid expensive
+copying on mutation. The `part` package comes with persistent `Map[K]V` and `Set[T]`.
+
+With this disclaimer out of the way, let's dive into how to use StateDB.
+
+## StateDB usage
+
+StateDB can be used directly as a normal library, or as a Hive Cell. For example
```
bimmlerd, line 52: nit: link `Hive` to github.com/cilium/hive

README.md
```diff
@@ -1,9 +1,509 @@
...
+usage as part of Hive, see `reconciler/example`. Here we show a standalone example.
```
bimmlerd, line 53: > standalone example

Additionally maybe a section on how to use it _with_ hive?

README.md
```diff
@@ -1,9 +1,509 @@
...
+For "very important objects", please consider storing an interface type instead that
+contains getter methods and a safe way of mutating the object, e.g. via the builder
+pattern or a constructor function. 
+
+Also prefer persistent/immutable data structures within the object to avoid expensive
+copying on mutation. The `part` package comes with persistent `Map[K]V` and `Set[T]`.
+
+With this disclaimer out of the way, let's dive into how to use StateDB.
+
+## StateDB usage
```
bimmlerd, line 50: Needs some structure, as pointed out. I think this document suffers a bit from trying to be all of reference documentation, a tutorial/get stared guide, and an overview.

In my mind, I'd organize it as follows: README is the overview - what is statedb, and maybe a quick, minimally complex example.

Then there's the bulk of this document, which I'd describe as a "guide" on how to use statedb. Not sure that wants to be in the README, but in any case it warrants a bit more structure in itself - headings to make it more "scannable". Could be it's own document, a bit like the guide to the hive (which, btw, we need to port to cilium/hive). Important part here is stuff like "use `Map[K, V] and Set[T]`" and why, and explanations of how to efficiently use statedb, why to use it in the first place.

finally, there's the reference docs, which is godocs and a pkg summary etc, maybe runnable examples. These will inevitably duplicate stuff from the README or the guide, but imo that's fine

bimmlerd (commented):
Great content, needs more structure. Couple of nits in some of the code bits. In terms of how to proceed- if you have time to organize the content a bit more, great, otherwise let's get this merged and then improve incrementally from there.

db.go
```diff
@@ -174,15 +191,20 @@ func (db *DB) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
 	return db.defaultHandle.WriteTxn(table, tables...)
 }
 
-func (db *DB) Start(cell.HookContext) error {
+// Start the background workers for the database.
+//
+// This starts the graveyard worker that deals with garbage collecting
+// deleted objects that are no longer necessary for Changes().
+func (db *DB) Start() error {
```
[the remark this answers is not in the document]
joamaki: Start() wasn't using the start context so dropped it. The context it created had a different lifecycle than the passed in context (e.g. it lived until Stop() and not only for the duration of Start()). The Start() and Stop() are mostly meant to be used as the Hive's start&stop hooks so I wouldn't change them semantically.

README.md
joamaki: Totally agree! In my mind I was writing a document akin to "Guide to the Hive" and not an overview as that's what I thought was missing the most. We could perhaps rename this to GUIDE.md and have an overview in README.md?

part/set.go
```diff
@@ -0,0 +1,151 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package part
+
+type Set[T any] struct {
```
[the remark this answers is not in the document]
joamaki: Added comment. 

Hmm not sure how useful stating the (vorst case) algorithmic complexity here is. It can in worst case be `O(k)` as radix trees devolve into essentially linked lists for degenerate cases like storing "aaaa" ... "zzzz": you'd have nodes a->a->a->a, but of course in practice it'd be sparse and more like `O(k log n)`.

README.md
joamaki: I'll leave that to another day.

db.go
```diff
@@ -174,15 +191,20 @@ func (db *DB) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
...
 	db.gcTrigger = make(chan struct{}, 1)
 	db.gcExited = make(chan struct{})
 	db.ctx, db.cancel = context.WithCancel(context.Background())
 	go graveyardWorker(db, db.ctx, db.gcRateLimitInterval)
 	return nil
 }
 
-func (db *DB) Stop(stopCtx cell.HookContext) error {
+// Stop the background workers.
+func (db *DB) Stop(stopCtx context.Context) error {
```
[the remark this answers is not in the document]
joamaki: I dropped the stop context for now.

part/set.go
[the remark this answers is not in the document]
bimmlerd: Yeah, that's a fair point. Good enough as is!

bimmlerd (approved):
Cool!

PR #26, opened 2024-05-08
Rename First to Get and Get to List/ListWatch

joamaki:
The First() method name has been an eye sore from the start. This is now a good point to rename it as we're about to switch Cilium to using cilium/statedb and need to refactor a bunch of code anyway.

Before:
```
  obj, rev, found := table.First(txn, MyIndex.Query("foo"))
  obj, rev, found, watch := table.FirstWatch(txn, MyIndex.Query("foo"))

  iter, watch := table.Get(txn, MyIndex.Query("foo"))
```

Now:
```
  obj, rev, found := table.Get(txn, MyIndex.Query("foo"))
  obj, rev, found, watch := table.GetWatch(txn, MyIndex.Query("foo"))

  iter := table.List(txn, MyIndex.Query("foo"))
  iter, watch := table.ListWatch(txn, MyIndex.Query("foo"))
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,14 +89,14 @@ func example() {
   myObjects.Insert(wtxn, &MyObject{3, "c"})
 
   // Modify an object
-  if obj, _, found := myObjects.First(wtxn, IDIndex.Query(1)); found {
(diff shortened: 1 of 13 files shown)


bimmlerd (approved):
lgtm

PR #27, opened 2024-05-08
reconciler: Make the Status.String() prettier

joamaki:
Before: "done (updated: 4.734193304s ago)"
After: "Done (4.7s ago)"


bimmlerd (approved):
so pretty

PR #29, opened 2024-05-23
part: Map/Set zero values and JSON marshalling

joamaki:
While trying to use part.Map and part.Set in some real-world code I quickly realized it was missing support for JSON marshalling and unmarshalling. To implement this we need a way to implement `func (m *Map[K, V) UnmarshalJSON(data []byte) error` and for that to work the zero value has to be usable. To get around this and in general make these a bit easier to use this PR moves the "toBytes" into a global registry rather than passing these functions in with `NewMap` and so on. **Is this easy enough to use?**, e.g. to use part.Map with a new key type one does this:
```
type MyKey string
func init() {
  part.RegisterKeyType(func(k MyKey) []byte { return []byte(k) })
}
```
If this is not done, then e.g. `Map.Set` will panic telling the user to do this. In some sense this idea is similar to e.g. https://pkg.go.dev/encoding/gob#Register (or what k8s client-go does) in that in order to support marshalling one needs to tell it about the type.

I'm also dropping the "fromBytes" (`func([]byte) K`) function as it was rather inconvenient and inefficient for complex composite key types to implement and opted to make the iterator just return the raw key ([]byte). **Is this a reasonable trade-off?** The reason I'm doing this is because it's rather annoying writing this function for `loadbalancer.L3n4Addr`, and we have fair bit of cases where we want these sort of composite key types and I think on iteration we're far more likely interested in the value anyway (the key often is within the value anyway).
EDIT: Hmm, or maybe we'll just store `K` along with the value. Taking a look at that.
EDIT2: Ok done this, now the `Map`  stores `(key, value)` pairs and the JSON encoding is `[(k,v), ...]`.

### DIFF
--- a/part/registry.go
+++ b/part/registry.go
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package part
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+	"reflect"
+	"sync"
+	"unicode/utf8"
+)
+
+// keyTypeRegistry is a registry of functions to convert to/from keys (of type K).
+// This mechanism enables use of zero value and JSON marshalling and unmarshalling
+// with Map and Set.
+var keyTypeRegistry sync.Map // map[reflect.Type]func(K) []byte
+
+// RegisterKeyType registers a new key type to be used with the Map and Set types.
+// Intended to be called from init() functions.
+// For Set-only usage only the [bytesFromKey] function is needed.
+func RegisterKeyType[K any](bytesFromKey func(K) []byte) {
+	keyType := reflect.TypeFor[K]()
+	keyTypeRegistry.Store(
+		keyType,
+		bytesFromKey,
+	)
+}
+
+func lookupKeyType[K any]() func(K) []byte {
+	keyType := reflect.TypeFor[K]()
+	funcAny, ok := keyTypeRegistry.Load(keyType)
+	if !ok {
+		panic(fmt.Sprintf("Key type %q not registered with part.RegisterMapKeyType()", keyType))
+	}
+	return funcAny.(func(K) []byte)
+}
+
+func init() {
+	// Register common key types.
+	RegisterKeyType[string](func(s string) []byte { return []byte(s) })
+	RegisterKeyType[[]byte](func(b []byte) []byte { return b })
+	RegisterKeyType[byte](func(b byte) []byte { return []byte{b} })
+	RegisterKeyType[rune](func(r rune) []byte { return utf8.AppendRune(nil, r) })
+	RegisterKeyType[complex128](func(c complex128) []byte {
+		buf := make([]byte, 0, 16)
+		buf = binary.BigEndian.AppendUint64(buf, math.Float64bits(real(c)))
+		buf = binary.BigEndian.AppendUint64(buf, math.Float64bits(imag(c)))
+		return buf
+	})
+	RegisterKeyType[float64](func(x float64) []byte { return binary.BigEndian.AppendUint64(nil, math.Float64bits(x)) })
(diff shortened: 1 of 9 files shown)


part/set.go
```diff
@@ -88,6 +77,12 @@ func (s Set[T]) All() SetIterator[T] {
 
 // Union combines the values in the two sets. Returns a new set.
 func (s Set[T]) Union(s2 Set[T]) Set[T] {
+	if s2.tree == nil {
+		return s
```
bimmlerd, line 81: Doesn't return a new set like the comment claims.

part/set.go
```diff
@@ -121,11 +116,41 @@ func (s Set[T]) Len() int {
 	return s.tree.size
 }
 
+// Equal returns true if the two sets contain the equal keys.
+func (s Set[T]) Equal(other Set[T]) bool {
+	switch {
+	case s.tree == nil:
+		return other.tree == nil
+	case other.tree == nil:
+		return s.tree == nil
```
bimmlerd, line 125: or just `false`.

part/set.go
```diff
@@ -141,6 +166,57 @@ func (s Set[T]) ToBytesFunc() func(T) []byte {
...
+	if d, ok := t.(json.Delim); !ok || d != '[' {
+		return fmt.Errorf("%T.UnmarshalJSON: expected '[' got %v", s, t)
+	}
+	for dec.More() {
+		var x T
+		err := dec.Decode(&x)
+		if err != nil {
+			return err
+		}
+		*s = s.Set(x)
```
bimmlerd, line 207: could be more efficient to gather the values and only then put them in a set; but not sure how much of a difference it makes

part/registry.go
```diff
@@ -0,0 +1,50 @@
...
+	RegisterKeyType[[]byte](func(b []byte) []byte { return b })
+	RegisterKeyType[byte](func(b byte) []byte { return []byte{b} })
+	RegisterKeyType[uint64](func(x uint64) []byte { return binary.BigEndian.AppendUint64(nil, x) })
+	RegisterKeyType[uint32](func(x uint32) []byte { return binary.BigEndian.AppendUint32(nil, x) })
+	RegisterKeyType[uint16](func(x uint16) []byte { return binary.BigEndian.AppendUint16(nil, x) })
+	RegisterKeyType[int64](func(x int64) []byte { return binary.BigEndian.AppendUint64(nil, uint64(x)) })
+	RegisterKeyType[int32](func(x int32) []byte { return binary.BigEndian.AppendUint32(nil, uint32(x)) })
+	RegisterKeyType[int16](func(x int16) []byte { return binary.BigEndian.AppendUint16(nil, uint16(x)) })
+	RegisterKeyType[int](func(x int) []byte { return binary.BigEndian.AppendUint64(nil, uint64(x)) })
+}
```
bimmlerd, line 50: I guess the builtin `bool`, `float` and `complex` types (eg https://pkg.go.dev/builtin#complex128) should be registered too? maybe also `error` and `rune`, even though I can't currently come up with super sensible use cases

bimmlerd (approved):
couple of nits but generally seems sensible

I think the register stuff is fine for custom types, but let's ensure all builtin types just work - I guess except channels/maps/slices.

part/set.go
joamaki: Reworded.

part/set.go
joamaki: No it needs to return true if both sets are empty, but I think I'll simplify this a bit.

part/set.go
joamaki: It would allocate twice as much memory as needed. However we can optimize this by using a transaction so it won't do unnecessary copies of the inner nodes.

part/registry.go
joamaki: hmm sure, we might as well. bool and error are definitely useful.

PR #31, opened 2024-05-31
reconciler: Add test case for the example

joamaki:
Add a simple test-case for the reconciler example to make sure it doesn't accidentally break in the future.

### DIFF
--- a/reconciler/example/test.sh
+++ b/reconciler/example/test.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR
+
+exec 3>&1
+exec &> test.log
+
+set -eux
+
+go build .
+./example &
+
+PID=$!
+cleanup() {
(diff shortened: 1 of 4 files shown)


reconciler/example/test.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
bimmlerd, line 2: hm, not sure I'm super excited about a bash test case that doesn't run on `go test ./...`

What was the reasoning in not making this a Go test? ease of starting/stopping processes etc?

reconciler/example/test.sh
```diff
@@ -0,0 +1,57 @@
...
+  else
+    echo "PASS" >&3
+  fi
+  kill -9 $PID
+  exit $s
+}
+trap 'cleanup' SIGINT SIGTERM EXIT
+
+for i in 1 2 3 4 5; do
+  curl -q http://localhost:8080/statedb &>/dev/null && break
```
bimmlerd, line 30: should this error out if it doesn't manage to come up soon enough?

PR #32, opened 2024-06-03
reconciler: Split full reconciliation into prune and refresh

joamaki:
The full reconciliation combined both pruning and refreshing of objects into one which is not always desirable. Sometimes we only need periodic (or one-off) pruning without forced Update() of the objects or the other way around. Additionally doing the refresh of all objects in one go can be very costly and possibly disruptive to the normal reconciliation, e.g. new changes won't be reconciled until the full reconciliation has finished.

To address this, split the 'FullReconciliationInterval' config option into 'PruneInterval' and 'RefreshInterval' and rework the implementation to implement pruning and refreshing separately. The refreshing is now performed simply by marking the objects pending if they have not have their status changed since 'RefreshInterval'. The marking is also spread over a period of RefreshInterval/2 in order to perform the refreshes slowly over time.


reconciler/reconciler_test.go
```diff
@@ -267,40 +264,29 @@ func testReconciler(t *testing.T, batchOps bool) {
...
+		h.expectOp(opUpdate(ID_1))
+		h.expectStatus(ID_1, reconciler.StatusKindDone, "")
 
-	assert.Greater(t, getInt(expVarMetrics.IncrementalReconciliationCountVar.Get("test")), int64(0), "IncrementalReconciliationCount")
-	assert.Greater(t, getFloat(expVarMetrics.IncrementalReconciliationDurationVar.Get("test/update")), float64(0), "IncrementalReconciliationDuration/update")
-	assert.Greater(t, getFloat(expVarMetrics.IncrementalReconciliationDurationVar.Get("test/delete")), float64(0), "IncrementalReconciliationDuration/delete")
-	assert.Greater(t, getInt(expVarMetrics.IncrementalReconciliationTotalErrorsVar.Get("test")), int64(0), "IncrementalReconciliationTotalErrors")
-	assert.Equal(t, getInt(expVarMetrics.IncrementalReconciliationCurrentErrorsVar.Get("test")), int64(0), "IncrementalReconciliationCurrentErrors")
+		// With 1 second delay and 100ms refresh interval, we should see at least 5 refreshes.
+		time.Sleep(time.Second)
```
joamaki, line 286: Not great to have this sort of approximate and time-based test case. Suggestions on how do this better are welcome.
Validated with `go test -c -race` and `stress ./reconciler.test` and got `1m50s: 575 runs so far, 0 failures` so it shouldn't at least be flaky.

marseel: I'm a big fan of using FakeClock, something like that:
https://github.com/cilium/cilium/pull/32588/files#diff-35fa8fe44daee2f4803801edfd6b2533b74caf8958ed58b3763feca25919e5a5R27

joamaki: Thanks, I'll look into that!

bimmlerd (approved):
generally lgtm
I think if the refresh interval is low then creating a write transaction per object to refresh will become expensive, but then again that's sort of a misconfiguration in the first place

reconciler/reconciler.go
```diff
@@ -184,3 +157,76 @@ func (r *reconciler[Obj]) loop(ctx context.Context, health cell.Health) error {
...
+			}
+		}
+
+		if len(toRefresh) == 0 {
+			continue
+		}
+
+		// Evenly spread the refreshes over a period of RefreshInterval/2 to avoid
+		// disruptions to normal reconciliation.
+		markTicker := time.NewTicker(r.Config.RefreshInterval / 2 / time.Duration(len(toRefresh)))
```
marseel, line 212: We discussed this offline, what `RefreshInterval` really means (i. e. what is the expected time for actual refresh / worst case time), if we want to have  `RefreshInterval / 2` in this line or maybe just `RefreshInterval` etc.

How about a totally different design, a little bit less discrete compared to this one:
- we have one "consumer" that pulls requests from workqueue to refresh with expected qps of ~`Table Elements / RefreshInterval`, but checks if the object needs updating (if it wasn't update in mean-time) before marking it as pending.
- we have one "producer" that watches for Table changes and adds them to something like delaying workqueue: https://github.com/kubernetes/client-go/blob/master/util/workqueue/delaying_queue.go

This would mean that expected refresh time is essentially almost always smaller than `~RefreshInterval`, it would evenly spread updates over time, while also ensuring that work is not wasted in case the object gets updated in mean-time.

joamaki: That's somewhat similar to what I initially had in mind, e.g. have a priority queue key'd on when the object should refresh, but I ended up going for the simpler design. I can take a look at how this would look like. I don't want to use the client-go workqueue as there's users of StateDB that don't want to pull in all those extra dependencies, but I can reuse the `retryQueue` priority queue implementation for this.

reconciler/types.go
```diff
@@ -15,15 +15,15 @@ import (
...
+	// Prune triggers an immediate pruning regardless of [PruneInterval].
 	// Implemented as a select+send to a channel of size 1, so N concurrent
 	// calls of this method may result in less than N full reconciliations.
+	// This still requires the table to be fully initialized to have an effect.
 	//
 	// Primarily useful in tests, but may be of use when there's knowledge
 	// that something has gone wrong in the reconciliation target and full
 	// reconciliation is needed to recover.
-	TriggerFullReconciliation()
+	Prune()
```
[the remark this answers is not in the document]
joamaki: You can force that by just doing a WriteTxn and marking all as pending (or refreshing).

reconciler/reconciler_test.go
joamaki: Rewrote the test to not delay.

PR #33, opened 2024-06-04
reconciler: Add StatusSet and optimize for multiple reconcilers

joamaki:
In order to efficiently support multiple reconcilers per object we need to make the reconciler aware of whether the object's data changed versus whether the reconciliation status changed. The current implementation only checked whether the object has a new revision when committing the reconciliation status, but in the presence of multiple reconcilers per object we need to actually check if the object was asked to be re-reconciled.

This commit adds the notion of "pending id" to 'Status' which allows the reconciler to compare the id of the object it reconciled with the id of the current object in the database and update it if the ids still match, regardless of the revision.

To make it easier to implement reconcilers for an object where the set of reconcilers is unknown beforehand, implement 'StatusSet', a named set of statuses.

### DIFF
--- a/reconciler/incremental.go
+++ b/reconciler/incremental.go
@@ -5,6 +5,7 @@ package reconciler
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/cilium/hive/cell"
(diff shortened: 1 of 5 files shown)


joamaki:
drafted until https://github.com/cilium/statedb/pull/34 is merged.

dylandreimerink (approved):
Very nice! Looks good to me

PR #34, opened 2024-06-14
reconciler: Add support for defaults and fix health status

joamaki:
reconciler: Do not query for object when retrying

    There's no need to requery for the object when retrying as the entry will be
    cleared from the queue if it has changed.

    This will also retain the original status of the object which allows the
    operations to distinquish whether the Update() call is a retry of a normal
    update or a refresh.

    Signed-off-by: Jussi Maki <jussi.maki@isovalent.com>

reconciler: Add support for default values in config

    Allow leaving the retry backoff, pruning and refreshing intervals
    unspecified and using the default values for them.

    The defaults proposed are:
    - Refresh*: No refreshing by default.
    - PruneInterval: No pruning by default.
    - RetryBackoff*: Retry with exponential backoff from 100ms to 1 minute
    - IncrementalRoundSize: Process 1000 objects per reconciliation round
    - RateLimiter: Do at most 100 rounds per second (100*1000 = 100k/s at most)

    Change the pruning to prune immediately when the table becomes initialized
    and then onwards prune every PruneInterval.

    Signed-off-by: Jussi Maki <jussi.maki@isovalent.com>

reconciler: Fix health reporting of retries

    The objects waiting for retry were not properly reflected in the health
    status. A reconciliation round after the failures cleared the health status
    rather than reflected the number of objects still waiting for retry.

    Fix this by remembering the error in the retry queue items and returning 
    those. Also modify the metrics implementation to properly reflect the
    "new" and "current" (new + waiting to retry) errors.

    Signed-off-by: Jussi Maki <jussi.maki@isovalent.com>

reconciler: Functional options API

    To make it easier to extend the reconciler API to support new mandatory or
    optional options, switch away from the Config struct to functional options.

    This also makes it easier to provide default values where the default isn't
    a zero value.

    Signed-off-by: Jussi Maki <jussi.maki@isovalent.com>


joamaki:
Marking as draft. Missed a few things...

joamaki:
This was supposed to be just adding the defaults for some of the config options but I ended up finding bunch of issues and fixing those as well.

I am still wondering whether the `Config` struct is the way to go or whether we should have a more of a "functional options" style thing, e.g. 
```
func myConfig(foos Table[Foo]) reconciler.Config[Foo] {
  return reconciler.For(foos, newOps(), Foo.Clone, Foo.GetStatus, Foo.SetStatus).
     // these are optional:
     Refresh(time.Minute, rate.NewLimiter(...)).
     Prune(time.Hour).
     RetryBackoff(100*time.Millisecond, time.Minute).
     RoundRateLimit(rate.NewLimiter(...)     
```

joamaki:
> > I am still wondering whether the Config struct is the way to go or whether we should have a more of a "functional options" style thing, e.g.
> 
> Functional options generally provide a better maintainable API, but since we are the primary user of the library that isn't super important. The UX argument is also good. I don't think it would make things worse, just more code.

I would argue we should not think of ourselves us the primary user of the library. Almost rather the opposite: we want to hand this over to other users and not use this ourselves if we can help it (vs educating others to use it).

joamaki:
@dylandreimerink @bimmlerd could you review when you have a bit of spare time? Not huge hurry. Thinking about bumping StateDB with these changes in cilium/cilium at start of v1.17.

dylandreimerink (approved):

reconciler/reconciler_test.go
```diff
@@ -301,9 +305,9 @@ func testReconciler(t *testing.T, batchOps bool) {
 	})
 
 	runTest("refreshing",
-		func(cfg *reconciler.Config[*testObject]) {
-			cfg.PruneInterval = 0
-			cfg.RefreshInterval = 500 * time.Millisecond
+		[]reconciler.Option{
+			reconciler.WithPruning(0), // Disable
```
bimmlerd, line 309: This feels like suboptimal DX. If pruning is on by default, we could alias this to `reconciler.WithoutPruning()` which reads easier.

reconciler/builder.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func WithRetry(minBackoff, maxBackoff time.Duration) Option {
+	return func(opts *options) {
+		opts.RetryBackoffMinDuration = minBackoff
+		opts.RetryBackoffMaxDuration = maxBackoff
+	}
+}
+
+// WithRoundLimits optionsthe reconciliation round limits.
```
bimmlerd, line 120: nit: broken doc comment

reconciler/builder.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func WithRefreshing(interval time.Duration, limiter *rate.Limiter) Option {
+	return func(opts *options) {
+		opts.RefreshInterval = interval
+		opts.RefreshRateLimiter = limiter
+	}
+}
+
+func WithRetry(minBackoff, maxBackoff time.Duration) Option {
```
bimmlerd, line 113: Should the doc comments from the options move to these public functions, as these are the API now?

reconciler/incremental.go
```diff
@@ -242,12 +232,24 @@ func (round *incrementalRound[Obj]) commitStatus() {
...
+		_, _, err := round.table.CompareAndSwap(wtxn, result.rev,
+			round.config.SetObjectStatus(obj, status))
+
+		if result.err != nil && err == nil {
+			// Reconciliation of the object had failed and the status was updated
+			// successfully (object had not changed). Queue the retry for the object.
+			newRevision := round.table.Revision(wtxn)
+			round.retries.Add(result.original.(Obj), newRevision, false, result.err)
+		} else if result.err != nil && err != nil {
+			fmt.Printf("FAIL queued %v for retry due to %s ! %s\n", result.original, result.err, err)
```
bimmlerd, line 251: Should this be a log? or a panic?

bimmlerd (approved):
couple of nits, otherwise lgtm

reconciler/builder.go
joamaki: Yes! Moved.

reconciler/incremental.go
joamaki: Neither. It was debug logging that I forgot there. Removed!

reconciler/reconciler_test.go
joamaki: Agreed. Added `WithoutPruning`.

PR #37, opened 2024-07-03
part: Add QuickTests

joamaki:
Add generative QuickTests to test out different
operations on generated inputs.

Fix the bug with closing the watch channel when
deleting single element tree with a zero-sized key.

### DIFF
--- a/part/part_test.go
+++ b/part/part_test.go
@@ -594,6 +594,32 @@ func Test_prefix(t *testing.T) {
 	assert.False(t, ok)
 }
 
+func Test_deleteEmptyKey(t *testing.T) {
+	tree := New[string]()
+
+	_, _, tree = tree.Insert([]byte{}, "x")
+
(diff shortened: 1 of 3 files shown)


part/part_test.go
```diff
@@ -594,6 +594,36 @@ func Test_prefix(t *testing.T) {
...
+	v, watch, ok := tree.Get([]byte{})
+	assert.True(t, ok)
+	assert.Equal(t, "x", v)
+	select {
+	case <-watch:
+		t.Fatalf("channel closed")
+	default:
+	}
+
+	tree.PrintTree()
```
bimmlerd, line 611: do you want to keep the prints?

bimmlerd (approved):
nice, TIL about `testing/quick`!

part/part_test.go
joamaki: Ugh I've been forgetting to remove my debug prints a lot lately. Fixed.

PR #38, opened 2024-07-11
reconciler: Fix refresh timer calculation

joamaki:
The refresh loop was supposed to wait until the oldest object's UpdatedAt time exceeds the RefreshInterval. Unfortunately the calculation was reversed (`UpdatedSince - RefreshInterval` when it should've been `RefreshInterval - UpdatedSince`) and the refresh duration to wait became negative, which caused the refreshLoop to busy-loop looking for the next object to refresh. 

This did not affect correctness as the timer was only used to wait before checking for expired objects and thus was not catched.

Before: 
```
agent.datapath.sysctl                            job-refresh     OK      Refreshing in -2m39.041586521s ...    
```
(job-refresh busy-loop'd constantly updating the health and resetting the timer to a negative value which then immediately fired)

After:
```
agent.datapath.sysctl                            job-refresh      OK      Next refresh in 9m45.997072899s  ...
```

Refresh in action:
```
root@kind-worker:/home/cilium# cilium statedb sysctl -w=10ms
# Name                                      Value   Status
net.ipv4.ip_forward                         1       Done (9.9m ago)
net.ipv4.conf.all.forwarding                1       Done (9.9m ago)
...
net.ipv4.conf.cilium_vxlan.accept_local     1       Refreshing (0.3ms ago)
net.ipv4.conf.cilium_vxlan.send_redirects   0       Refreshing (0.3ms ago)
net.ipv6.conf.cilium_vxlan.forwarding       1       Refreshing (0.3ms ago)
net.ipv4.conf.cilium_vxlan.forwarding       1       Refreshing (0.3ms ago)
net.ipv4.conf.cilium_vxlan.rp_filter        0       Refreshing (0.3ms ago)
net.ipv4.conf.cilium_vxlan.send_redirects   0       Done (10.7ms ago)
net.ipv6.conf.cilium_vxlan.forwarding       1       Done (10.7ms ago)
net.ipv4.conf.cilium_vxlan.forwarding       1       Done (10.7ms ago)
...
```

### DIFF
--- a/reconciler/reconciler.go
+++ b/reconciler/reconciler.go
@@ -126,11 +126,9 @@ func (r *reconciler[Obj]) prune(ctx context.Context, txn statedb.ReadTxn) error 
 func (r *reconciler[Obj]) refreshLoop(ctx context.Context, health cell.Health) error {
 	lastRevision := statedb.Revision(0)
 
-	refreshTimer := time.NewTimer(r.config.RefreshInterval)
+	refreshTimer := time.NewTimer(0)
 	defer refreshTimer.Stop()
 
-	health.OK(fmt.Sprintf("Refreshing in %s", r.config.RefreshInterval))
-outer:
 	for {
 		// Wait until it's time to refresh.
 		select {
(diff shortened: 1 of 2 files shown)


reconciler/reconciler.go
```diff
@@ -140,27 +138,25 @@ outer:
...
-		// pending in order for them to be reconciled again.
+		// refresh in order for them to be reconciled again.
 		iter := r.config.Table.LowerBound(r.DB.ReadTxn(), statedb.ByRevision[Obj](lastRevision+1))
 		indexer := r.config.Table.PrimaryIndexer()
+		now := time.Now()
 		for obj, rev, ok := iter.Next(); ok; obj, rev, ok = iter.Next() {
 			status := r.config.GetObjectStatus(obj)
 
+			// The duration elapsed since this was updated.
+			updatedSince := now.Sub(status.UpdatedAt)
```
giorio94, line 153: I wonder whether it could be more appropriate to use `time.Since` here, considering that marking an object for refresh is a potentially slow operation due to rate limiting.

reconciler/reconciler.go
```diff
@@ -140,27 +138,25 @@ outer:
...
 		}
 
-		now := time.Now()
+		durationUntilRefresh := r.config.RefreshInterval
 
 		// Iterate over the objects in revision order, e.g. oldest modification first.
 		// We look for objects that are older than [RefreshInterval] and mark them for
-		// pending in order for them to be reconciled again.
+		// refresh in order for them to be reconciled again.
 		iter := r.config.Table.LowerBound(r.DB.ReadTxn(), statedb.ByRevision[Obj](lastRevision+1))
```
giorio94, line 146: Are revisions starting from one? Otherwise it seems to me that the first object is skipped.

giorio94 (approved):
A couple of comments inline for my own understanding. Otherwise looks good to me.

reconciler/reconciler.go
joamaki: Yes, revisions start from 1.

reconciler/reconciler.go
joamaki: Ah yes. Good catch!

giorio94 (approved):
Thanks!

PR #41, opened 2024-07-31
statedb: Add checks to see if table is registered

dylandreimerink:
If a user forgets to register a table with the DB instance and then proceeds to use it, the tablePos defaults to 0. This causes writes to the table to end up in the wrong table.

This commit defaults the table pos to -1 to cause deliberate panics if the code ever attempts to use the index of a unregistered table. This commit also adds checks on the read and write transactions to detect when an unregistered table is being used and throw a clear error.

### DIFF
--- a/db.go
+++ b/db.go
@@ -249,6 +249,9 @@ func (h Handle) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
 	smus := internal.SortableMutexes{}
 	for _, table := range allTables {
 		smus = append(smus, table.sortableMutex())
+		if table.tablePos() < 0 {
+			panic(tableError(table.Name(), ErrTableNotRegistered))
+		}
 	}
 	lockAt := time.Now()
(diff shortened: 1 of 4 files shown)


db.go
```diff
@@ -246,6 +246,11 @@ type Handle struct {
 func (h Handle) WriteTxn(table TableMeta, tables ...TableMeta) WriteTxn {
 	db := h.db
 	allTables := append(tables, table)
+	for _, table := range allTables {
+		if table.tablePos() < 0 {
+			panic(tableError(table.Name(), ErrTableNotRegistered))
```
joamaki, line 251: Can we move this into the loop below that gathers the mutexes?

joamaki (approved):

db.go
dylandreimerink: Yes, of course we can.

dylandreimerink:
Shall I merge the PR?

PR #44, opened 2024-08-16
Add Modify() method

joamaki:
This adds Modify() to `part` and to `RWTable[T]`. 

Modify() can be used in Cilium to efficiently implement merging of external data into internal objects
that carry state that we want to merge with rather than replace. A follow-up PR in cilium/cilium will
add the merging support to `pkg/k8s/statedb.go`.


dylandreimerink (approved):
Nice addition

PR #46, opened 2024-08-20
Fix comment posting on failed CI run and cleanup benchmarks

joamaki:
If a PR fails the failed "make test" output wasn't posted. Add a "if: always()" to the comment posting to always post it.
This was manually tested.

Fix the secondary index benchmark not to recreate the tags set on each insert as that's not what is important for the benchmark.
Additionally drop the very large batch sizes from benchmarks as those use up fair bit of memory and don't provide a useful signal.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -36,6 +36,7 @@ jobs:
           echo '```' >> results.comment
           
       - name: results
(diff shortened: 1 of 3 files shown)


dylandreimerink (approved):
LGTM

PR #47, opened 2024-08-20
reconciler: Prune after initialized even with empty table

joamaki:
Pruning should be performed when the table has initialized even when the table is empty. To implement this a watch channel is added for initialization and Initialized() now returns (bool, <-chan struct{}).
The reconciler can now wait for initialization and immediately trigger the pruning.

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -941,36 +941,58 @@ func TestDB_Initialization(t *testing.T) {
 
 	db, table, _ := newTestDB(t, tagsIndex)
 
+	// Using Initialized() before any initializers are registered
+	// will return true and a closed channel.
+	init, initWatch := table.Initialized(db.ReadTxn())
+	require.True(t, init, "Initialized should be true")
+	select {
+	case <-initWatch:
+	default:
+		t.Fatalf("Initialized watch channel should be closed")
+	}
+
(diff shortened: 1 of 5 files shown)


dylandreimerink:
Wouldn't it make more sense to change stateDB to invalidate all watch channels of a table when that table goes from uninitialized to initialized, even if its empty? Seems like a more elegant solution. Also for other users of stateDB who might await initialization of one or more tables before starting processing:
```
for {
  rxn := db.ReadTxn()
  if tbl.Initialized(rxn) {
    break
  }
  _, watch := tbl.AllWatch()
  watch->
}
```

joamaki:
> Wouldn't it make more sense to change stateDB to invalidate all watch channels of a table when that table goes from uninitialized to initialized, even if its empty? Seems like a more elegant solution. Also for other users of stateDB who might await initialization of one or more tables before starting processing:
> 
> ```
> for {
>   rxn := db.ReadTxn()
>   if tbl.Initialized(rxn) {
>     break
>   }
>   _, watch := tbl.AllWatch()
>   watch->
> }
> ```

"invalidate all watch channels" is really non-trivial, likely highly invasive and I don't know how to do it.

dylandreimerink (approved):
:rocket:

PR #48, opened 2024-09-02
http: /changes API for streaming changes

joamaki:
This allows streaming changes (upserts and deletes) over the HTTP API.

Example usage:
```
  examples := statedb.NewRemoteTable[*Example](url, "examples")

  // Stream changes until [ctx] is cancelled.
  changes, errChan := examples.Changes(ctx)
  for change, rev, ok := changes.Next(); ok; change, rev, ok = changes.Next() {
    example := change.Object
    if change.Deleted {
      // deleted
    } else {
      // upserted
    }
  }
  err := <-errChan
```

### DIFF
--- a/http_client.go
+++ b/http_client.go
@@ -9,6 +9,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -77,13 +78,6 @@ func (t *RemoteTable[Obj]) query(ctx context.Context, lowerBound bool, q Query[O
 	}
 	return &remoteGetIterator[Obj]{json.NewDecoder(resp.Body), errChanSend}, errChan
 }
-func (t *RemoteTable[Obj]) Get(ctx context.Context, q Query[Obj]) (Iterator[Obj], <-chan error) {
-	return t.query(ctx, false, q)
-}
-
-func (t *RemoteTable[Obj]) LowerBound(ctx context.Context, q Query[Obj]) (Iterator[Obj], <-chan error) {
-	return t.query(ctx, true, q)
-}
 
 type remoteGetIterator[Obj any] struct {
 	decoder *json.Decoder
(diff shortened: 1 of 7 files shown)


joamaki:
With this new API we can change "cilium-dbg statedb <table> --watch" to also show deletions. PR to prepare that: https://github.com/cilium/cilium/pull/34635.

http_client.go
```diff
@@ -125,3 +119,69 @@ func (it *remoteGetIterator[Obj]) Next() (obj Obj, revision Revision, ok bool) {
...
+			// Keep-alive message, skip it.
+			continue
+		}
+
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				close(it.errChan)
+				return
+			}
+			it.decoder = nil
```
pippolo84, line 177: Since the decoder is set to `nil` in case of error, shouldn't we close `it.errChan` in that case too? Otherwise we'll leak the channel, as we won't have a chance to reach line 173 anymore.

pippolo84 (commented):
Overall LGTM, just a question left inline

http_client.go
```diff
@@ -125,3 +119,69 @@ func (it *remoteGetIterator[Obj]) Next() (obj Obj, revision Revision, ok bool) {
...
+
+func (t *RemoteTable[Obj]) Changes(ctx context.Context) (iter Iterator[Change[Obj]], errChan <-chan error) {
+	// Use a channel to return errors so we can use the same Iterator[Obj] interface as StateDB does.
+	errChanSend := make(chan error, 1)
+	errChan = errChanSend
+
+	url := t.base.JoinPath("/changes", t.tableName)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url.String(), nil)
+	if err != nil {
+		fmt.Printf("err: %s\n", err)
```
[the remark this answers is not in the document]
joamaki: Oops, removed.

http_client.go
joamaki: Ah good point. There were multiple places where errChan should've been closed. And it makes more sense to send a `nil` error to the chan and then close. Fixed!

PR #49, opened 2024-09-02
Fix ChangeIterator watch channel on partial iteration

joamaki:
Add a regression test for Changes() API to show the issues and fix the issues:

1) If the ChangeIterator is not exhausted always return a closed watch channel to trigger the iteration.

2) The initial "iter.watch" channel does not need to be closed, but should rather be a watch channel against the revision index to avoid unnecessary refresh.

### DIFF
--- a/regression_test.go
+++ b/regression_test.go
@@ -5,6 +5,7 @@ package statedb
 
 import (
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


regression_test.go
```diff
@@ -72,5 +73,79 @@ func Test_Regression_29324(t *testing.T) {
...
+	changes, err := table.Changes(wtxn)
+	require.NoError(t, err, "Changes")
+	wtxn.Commit()
+
+	n := 0
+	_, _, ok := changes.Next()
+	require.False(t, ok, "expected no changes")
+
+	// On first call to Watch() the iterator is refreshed and a closed watch channel
+	// is returned.
```
pippolo84, line 95: If I get this right, this comment is not true anymore now that the initial watch channel returned is not a `closedWatchChannel`. Should we update this in the last commit to avoid confusion?

pippolo84 (approved):
LGTM apart from a possibly confusing comment in the regression test (see comment inline) :white_check_mark:

regression_test.go
joamaki: It is still true. `Watch` returns a closed watch channel when the iterator has not been consumed.

PR #50, opened 2024-09-05
Use Go standard library slices package instead of x/exp/slices

tklauser:
The former is available since Go 1.21 and the latter is merely a wrapper around it when using Go ≥ 1.21. Use the standard library package directly.

We cannot yet switch out `x/exp/maps` for the standard library `maps` package as the `maps.Keys` function used in `statedb` code base is only available in Go ≥ 1.23.

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -7,6 +7,7 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"maps"
 	"math/rand"
 	"os"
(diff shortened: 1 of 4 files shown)


pippolo84 (approved):

joamaki:
Just merged the range-funcs PR, so cilium/statedb is now at Go v1.23. Could you rebase and switch the maps package too?

PR #51, opened 2024-09-06
Unify DB and Handle types into just DB

joamaki:
This removes the separate `Handle` type to allow decorating `DB` with the handle name in Hive:
```
       hive.Options{
               // Create a named DB handle for each module.
               ModuleDecorator: func(db *statedb.DB, id cell.ModuleID) *statedb.DB {
                       return db.NewHandle(string(id))
               },
       },
```

With this we can get StateDB metrics accounted by module, allowing us to for example seeing which module
has long-running `WriteTxn`s etc.

### DIFF
--- a/db.go
+++ b/db.go
@@ -82,6 +82,12 @@ import (
 //  6. Periodically garbage collect the graveyard by finding
 //     the lowest revision of all delete trackers.
 type DB struct {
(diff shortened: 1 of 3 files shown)


pippolo84 (approved):
:100:

PR #52, opened 2024-09-06
Validate the table name in NewTable

joamaki:
To avoid constructing table names which are not usable for metrics (e.g. as Prometheus metrics label), restrict the allowed name to "^[a-z][a-z0-9_\\-]{1,30}$".


README.md
```diff
@@ -239,12 +239,15 @@ func NewMyObjectTable() (statedb.RWTable[*MyObject], error) {
...
 
 The `NewTable` function takes the name of the table, a primary index and zero or
-more secondary indexes. It returns a `RWTable`, which is an interface for both
-reading and writing to a table. An `RWTable` is a superset of `Table`, an interface
-that contains methods just for reading. This provides a simple form of type-level
-access control to the table. `NewTable` may return an error if the indexers are
-malformed, for example if `IDIndex` is not unique (primary index has to be), or if
-the indexers have overlapping names.
+more secondary indexes. The table name must match the regular expression
+"^[a-z][a-z0-9_\\-]{0,30}$".
```
ovidiutirla, line 243: Should we allow `_` or should we enforce `-` to be used always? 

Is there a particular reason for selecting 30? 30 chars -> ~4-8 words.
Considering a metric name needs to be usually below 200 chars, but the table name could be used in situation like jobgroup prefix -> table name -> suffix, I think it's more than enough.

The largest table name we currently have is 16 chars, `bandwidth-qdiscs` which IMO is fine.

ovidiutirla (approved):
LGTM.

TODO, I think we need to put a validation on the jobgroups too, they also emit metrics.

README.md
joamaki: There's two parts to the names in the `job` library. One is the name used in the module health and that is concat of the full module id (`agent.control-plane.foo`) and the job name (`job-control-loop`). The module ids are already validated, but indeed we don't validate the job name (e.g. the `name` argument passed to `job.OneShot` for example). 

Would you be interested in doing a PR for `cilium/hive` to add this validation? Probably same or similar constraint as used here makes sense.

joamaki: > Should we allow _ or should we enforce - to be used always?

I think allowing both is fine and using a similar constraint as we use in e.g. `cilium/hive` for module ids makes it easier to guess what is allowed.

PR #54, opened 2024-09-17
testutils: Add script-based test utilities

joamaki:
Add testutils with functions for creating commands for [testscript](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript).

This allows implementing tests as scripts, which becomes useful when tests perform multiple steps and need to verify one or more tables on each step. This aims to be a better way of writing test-suites that work against a set of input files and golden output files.

Each test suite would define their own set of commands with some table-based ones built using the utils and perhaps some specific to the thing being tested.

testscript docs: https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript
blog posts for better intro: https://encore.dev/blog/testscript-hidden-testing-gem, https://atlasgo.io/blog/2024/09/09/how-go-tests-go-test

and the reason I didn't use rsc.io/script: https://github.com/rsc/script/issues/4


testutils/script.go
```diff
@@ -0,0 +1,164 @@
...
+		lines := strings.Split(data, "\n")
+		if len(lines) < 1 {
+			ts.Fatalf("input too short")
+		}
+		lines = slices.DeleteFunc(lines, func(line string) bool {
+			return strings.TrimSpace(line) == ""
+		})
+
+		header := zero.TableHeader()
+		columnNames := strings.Split(lines[0], "\t")
```
joamaki, line 57: TODO: use strings.Fields so table can be nicely formatted in txtar.

testutils/script.go
```diff
@@ -0,0 +1,206 @@
...
+	"time"
+
+	"github.com/cilium/statedb"
+	"github.com/rogpeppe/go-internal/testscript"
+	"gopkg.in/yaml.v3"
+)
+
+type Cmd = func(ts *testscript.TestScript, neg bool, args []string)
+
+func ShowTableCmd[Obj statedb.TableWritable](db *statedb.DB, tbl statedb.Table[Obj]) Cmd {
```
joamaki, line 22: Hmm, I think I'll try another variation to add completely generic statedb commands that just need `DB`, e.g.
"cmp_table services", "show_table services", "tables" (like sqlite .tables), etc.

joamaki: Done.

testutils/script.go
```diff
@@ -0,0 +1,499 @@
...
+			ts.Logf("%s", buf.String())
+		} else if err := os.WriteFile(ts.MkAbs(*file), buf.Bytes(), 0644); err != nil {
+			ts.Fatalf("WriteFile(%s): %s", *file, err)
+		}
+	}
+}
+
+func CompareTableCmd(ts *testscript.TestScript, neg bool, args []string) {
+	var flags flag.FlagSet
+	wait := flags.Duration("wait", 0, "Amount of time to wait for the table contents to match")
```
joamaki, line 175: Don't like naming this `-wait`. `-retry`? `-eventually` (too long?)?

types.go
```diff
@@ -226,6 +230,8 @@ type TableMeta interface {
 	secondary() map[string]anyIndexer      // Secondary indexers (if any)
 	sortableMutex() internal.SortableMutex // The sortable mutex for locking the table for writing
 	anyChanges(txn WriteTxn) (anyChangeIterator, error)
+	proto() any
+	unmarshalYAML(data string) (any, error)
```
ovidiutirla, line 234: should we add this here, part of the TableMeta as it's only usage will be part of this `script.go`, or extract to a different package/even the scriptutils.

testutils/script.go
```diff
@@ -0,0 +1,499 @@
...
+
+func ShowTableCmd(ts *testscript.TestScript, neg bool, args []string) {
+	db := getDB(ts)
+	if len(args) != 1 {
+		ts.Fatalf("usage: show_table <table>")
+	}
+	ts.Logf("%s", showTable(ts, db, args[0]).String())
+}
+
+func WriteTableCmd(ts *testscript.TestScript, neg bool, args []string) {
```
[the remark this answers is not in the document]
joamaki: For negating the result of the command. Depends on the command itself on what the meaning is.
Docs has more about it: https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript

testutils/script.go
joamaki: I don't think we care about exponential backoff in tests that much. I'll make the default that it waits some amount of time and make this `-timeout`.

testutils/script.go
```diff
@@ -0,0 +1,499 @@
...
+			lines = lines[1:]
+		}
+		for _, line := range lines {
+			fmt.Fprintf(w, "+ %s\n", line)
+			equal = false
+		}
+		if equal {
+			return
+		}
+		w.Flush()
```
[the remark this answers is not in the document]
joamaki: No because we don't care about the diff if it's equal.

types.go
joamaki: Yeah, I'll need to revisit these weird ones a bit and see how I can mostly hide them. They do need to be in the `genTable` though as that's the place where we know what `Obj` is.

This one was a bit annoying as I need to call `yaml.Unmarshal` with a concrete thing and if it goes via the `any` it'll just end up producing `map[string]any` as a result. Will have to play around with it.

testutils/script.go
```diff
@@ -0,0 +1,499 @@
...
+	db := getDB(ts)
+	if len(args) != 1 {
+		ts.Fatalf("usage: show_table <table>")
+	}
+	ts.Logf("%s", showTable(ts, db, args[0]).String())
+}
+
+func WriteTableCmd(ts *testscript.TestScript, neg bool, args []string) {
+	if len(args) < 1 || len(args) > 5 {
+		ts.Fatalf("usage: write_table <table> (-to=<file>) (-columns=<col1,col2,...>) (-format={table*,yaml})")
```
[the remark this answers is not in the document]
ovidiutirla: I think it's a bit weird to have long names with one `-`, so what do you think about renaming them?
like:

`-to` -> `-d` (destination)
`-columns` -> `-c`
`-format` -> `-f`

etc?

testutils/script.go
```diff
@@ -0,0 +1,499 @@
...
+
+	// Sort the args to allow the table name at any position.
+	slices.SortFunc(args, func(a, b string) int {
+		switch {
+		case a[0] == '-':
+			return 1
+		case b[0] == '-':
+			return -1
+		default:
+			return cmp.Compare(a, b)
```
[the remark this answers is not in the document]
joamaki: Yeah, that's not supported. Table names cannot contain spaces. And yeah, column names might have spaces, but that breaks other stuff so one shouldn't do that.

testutils/script.go
```diff
@@ -0,0 +1,499 @@
...
+		if len(*columns) > 0 {
+			ts.Fatalf("-columns not supported with -format=yaml/json")
+		}
+
+		txn := db.ReadTxn()
+		meta := db.GetTable(txn, tableName)
+		if meta == nil {
+			ts.Fatalf("table %q not found", tableName)
+		}
+		tbl := statedb.AnyTable{Meta: meta}
```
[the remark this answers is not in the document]
joamaki: yes, fixing :D

testutils/script.go
```diff
@@ -0,0 +1,499 @@
...
+	for {
+		lines = origLines
+
+		// Create the diff between 'lines' and the rows in the table.
+		equal := true
+		var diff bytes.Buffer
+		w := tabwriter.NewWriter(&diff, 5, 4, 3, ' ', 0)
+		fmt.Fprintf(w, "  %s\n", strings.Join(columnNames, "\t"))
+
+		txn = db.ReadTxn()
```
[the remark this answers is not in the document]
joamaki: Oops looks like a leftover. Removed.

testutils/script.go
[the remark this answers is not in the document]
joamaki: That makes it harder to read so not a fan of that.

types.go
joamaki: Yeah I don't think there's a way around having `unmarshalYAML` here. We need to call `yaml.Unmarshal` on the concrete type as `Unmarshal` reflects on the type of the destination value and if we go via `any` we lose the type information at runtime. What I'll do though is split of the weird internal things into `tableInternal` interface so it's not so visible.

testutils/script.go
```diff
@@ -0,0 +1,496 @@
...
+		equal := true
+		var diff bytes.Buffer
+		w := tabwriter.NewWriter(&diff, 5, 4, 3, ' ', 0)
+		fmt.Fprintf(w, "  %s\n", joinByPositions(columnNames, columnPositions))
+
+		for obj := range tbl.All(db.ReadTxn()) {
+			rowRaw := takeColumns(obj.(statedb.TableWritable).TableRow(), columnIndexes)
+			row := joinByPositions(rowRaw, columnPositions)
+			if grepRe != nil {
+				if !grepRe.Match([]byte(row)) {
```
ovidiutirla, line 230: should we join these two ifs?

ovidiutirla (approved):
LGTM

PR #57, opened 2024-10-01
Reimplement script commands against hive script

joamaki:
Reimplement the script commands with the hive/script library. And with these reimplement the reconciler tests. I tried to be faithful to the original test cases, but please double-check.

See [testdata/db.txtar](57/files#diff-d53c2d912df4886bbd882e3ba324bb33ebcdabe7c3cfac50efcde2b81f9800c0) for an overview of the commands.

I'll do a separate pass to add documentation to the README here and to the cilium docs once we've gotten a bit more experience with this.

```
❯ go run .
time=2024-10-09T15:10:36.412+02:00 level=INFO msg=Invoked duration=121.02µs function="main.writeStuff (.../statedb/scriptexample/main.go:75)"
> hive start
time=2024-10-09T15:10:37.713+02:00 level=INFO msg=Starting
time=2024-10-09T15:10:37.713+02:00 level=INFO msg="Start hook executed" function="statedb.newHiveDB.func1 (.../cilium/statedb/cell.go:34) (statedb)" duration=74.336µs
time=2024-10-09T15:10:37.713+02:00 level=INFO msg="Start hook executed" function="*job.group.Start (example)" duration=14.949µs
time=2024-10-09T15:10:37.713+02:00 level=INFO msg=Started duration=502.237µs
> db tables
Name   Object count   Deleted objects   Indexes   Initializers   Go type      Last WriteTxn
ex     3              0                 id, x     []             main.ExObj   example (87.0ms ago, locked for 148.4us)
> db show ex
[stdout]
Name    X
count   3
one     1
two     2
> db list ex one
[stdout]
Name   X
one    1
> db prefix ex o
[stdout]
Name   X
one    1
> db
<stdin>:0: db: expected command (show, cmp, delete, prefix, lowerbound, tables, insert, get, list, initialized)
> db initialized
ex initialized
> db prefix -delete ex o
Deleted 1 objects
[stdout]
...
```


script.go
```diff
@@ -0,0 +1,581 @@
...
+		w = f
+	}
+
+	tw := tabwriter.NewWriter(w, 5, 4, 3, ' ', 0)
+	header := tbl.TableHeader()
+	fmt.Fprintf(tw, "%s\n", strings.Join(header, "\t"))
+
+	var it iter.Seq2[any, uint64]
+	if lowerbound {
+		it = tbl.LowerBound(txn, args[1])
```
joamaki, line 440: The querying needs a bit more work. Should add support for specifying which index to query and perhaps extend the `Index` type to have a `FromString` method (`func(string) index.Key`) so this works for any type of key and index that implements it.

script.go
```diff
@@ -0,0 +1,581 @@
...
+func ShowCmd(db *DB) script.Cmd {
+	return script.Command(
+		script.CmdUsage{
+			Summary: "Show StateDB table",
+			Args:    "table",
+		},
+		func(s *script.State, args ...string) (script.WaitFunc, error) {
+			if len(args) != 1 {
+				return nil, fmt.Errorf("expected table name")
+			}
```
joamaki, line 100: Would be good to support watching (e.g. use `Changes()`, show deletes as striked out lines etc.).

script.go
```diff
@@ -0,0 +1,581 @@
...
+	return script.Command(
+		script.CmdUsage{
+			Summary: "Show StateDB tables",
+			Args:    "table",
+		},
+		func(s *script.State, args ...string) (script.WaitFunc, error) {
+			txn := db.ReadTxn()
+			tbls := db.GetTables(txn)
+			w := tabwriter.NewWriter(s.LogWriter(), 5, 4, 3, ' ', 0)
+			fmt.Fprintf(w, "Name\tObject count\tGraveyard count\tIndexes\tInitializers\tGo type\n")
```
joamaki, line 79: We could fairly easily keep some overall stats for each table and show them here (read and write rates etc.).
Also recording who has the current WriteTxn (lock) on the table and since when is easy to add.

table.go
```diff
@@ -129,6 +159,27 @@ type genTable[Obj any] struct {
 	primaryAnyIndexer    anyIndexer
 	secondaryAnyIndexers map[string]anyIndexer
 	indexPositions       map[string]int
+	acquiredInfo         atomic.Pointer[acquiredInfo]
+}
+
+func (t *genTable[Obj]) acquired(handle string, acquiredAt time.Time) {
+	t.acquiredInfo.Store(&acquiredInfo{
```
joamaki, line 166: This adds an extra allocation to the `WriteTxn` call per table. It doesn't seem to have a huge difference in benchmarks for 1 object per WriteTxn (few % slower) so I think it's worth the trade-off as the information this gives is quite useful.

joamaki:
```
Name   Object count   Deleted objects   Indexes   Initializers   Go type      Last WriteTxn
ex     2              0                 id, x     [foo]          main.ExObj   example (1.2s ago, locked for 10.0us)
```
(the "Last WriteTxn" column is derived from this)

joamaki: Reimplemented this by storing the data in `txn` instead. Benchmarks back to previous numbers now.

index/string.go
```diff
@@ -12,6 +12,10 @@ func String(s string) Key {
 	return []byte(s)
 }
 
+func FromString(s string) (Key, error) {
```
joamaki, line 15: I think it makes sense to have also these utility functions here for the `FromString` index field. Not sure what the naming convention should be.
`index.Int64`, `index.Int64FromString`? `index.Int64String`?

part/set.go
```diff
@@ -208,6 +211,32 @@ func (s *Set[T]) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (s Set[T]) MarshalYAML() (any, error) {
```
joamaki, line 214: Oh, I might need `part.Map` yaml marshalling as well. Taking a look...

types.go
```diff
@@ -282,10 +287,26 @@ func ByRevision[Obj any](rev uint64) Query[Obj] {
...
-	Unique     bool
+
+	// FromKey converts the index key into a raw key.
+	// With this we can perform Query() against this index with
+	// the [Key] type.
+	FromKey func(key Key) index.Key
+
+	// FromString is an optional conversion from string to a raw key.
+	// If implemented allows script commands to query with this index.
+	FromString func(key string) (index.Key, error)
```
joamaki, line 304: I wonder if we want to make this mandatory or not. Would be a breaking change if we require it.
Might also make sense to have constructors for `Index` instead of having to fill in the struct.
I'm inclined to not make this mandatory as of yet as I want these changes to be non-breaking for now.

script.go
```diff
@@ -0,0 +1,682 @@
...
+		}
+	}
+}
+
+func writeObjects(tbl *AnyTable, it iter.Seq2[any, Revision], w io.Writer, columns []string, format string) error {
+	if len(columns) > 0 && format != "table" {
+		return fmt.Errorf("-columns not supported with non-table formats")
+	}
+	switch format {
+	case "yaml":
```
joamaki, line 524: Kind of tempted to pull in https://github.com/itchyny/gojq into this to allow assertions with a `jq` query. Otoh, it's a pretty big dependency so perhaps best to keep this minimal until there's a real need.

script.go
```diff
@@ -0,0 +1,682 @@
...
+			}
+			if _, err := w.Write(out); err != nil {
+				return err
+			}
+		}
+		return nil
+	case "table":
+		header := tbl.TableHeader()
+		if header == nil {
+			return fmt.Errorf("objects in table %q not TableWritable", tbl.Meta.Name())
```
joamaki, line 563: Here's perhaps a good reason to change `NewTable[Obj any]` into `NewTable[Obj TableWritable]` (or maybe make a `TableObject` constraint). Again don't want to make this into a breaking change yet, but maybe for statedb v0.4...

part/set.go
joamaki: Done.

index/string.go
joamaki: Added the `*String` variants in the last commit, e.g. `index.Int32String` and so on.

pippolo84 (commented):
Did a first pass and left two minor nits, gonna do another pass later focusing on reconciler tests.

script.go
```diff
@@ -0,0 +1,682 @@
...
+			Args:    "(-o=<file>) (-columns=col1,...) (-format={table,yaml,json}) table",
+		},
+		func(s *script.State, args ...string) (script.WaitFunc, error) {
+			flags := newCmdFlagSet()
+			file := flags.String("o", "", "File to write to instead of stdout")
+			columns := flags.String("columns", "", "Comma-separated list of columns to write")
+			var cols []string
+			if len(*columns) > 0 {
+				cols = strings.Split(*columns, ",")
+			}
```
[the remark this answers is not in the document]
joamaki: oopsie, indeed! I'll add tests to validate this as well.

index/string.go
pippolo84: > Or should I add it for consistency?

As a personal preference I'd go for `index.StringString` to be consistent, but agree that it is not a nice name :sweat:
No strong opinion on this, in the end both are fine for me.

pippolo84 (approved):
This is great! :rocket: 

To the best of my knowledge, the new script-based tests seem reasonably equivalent to the old ones :+1: 
Tried also to `stress` the new test and it behaves correctly after more than 10k runs. :tada:

PR #59, opened 2024-10-03
Fix Prefix and LowerBound on non-unique indexes

joamaki:
Prefix and LowerBound searches did not properly deal with non-unique indexes. In these indexes the keys are encoded as <secondary><primary><secondary len>, and prefix searching needs to make sure the secondary key length on the results is equal or longer than the search key as otherwise we might match into the primary key.

For example if the object is struct{A, B string}, with A being primary and we have {"a", "a"}, then the secondary index is key'd as "aa<1>". A prefix search with "aa" must not match since the secondary index contains only an object with key "a".

Fix this by using a special iteration on non-unique indexes that checks the length of the secondary key and ignores any other matches that are due to matching into the primary key.

### DIFF
--- a/iterator.go
+++ b/iterator.go
@@ -4,6 +4,7 @@
 package statedb
 
 import (
+	"bytes"
 	"fmt"
 	"iter"
 	"slices"
@@ -71,29 +72,11 @@ func partSeq[Obj any](iter *part.Iterator[object]) iter.Seq2[Obj, Revision] {
 	}
 }
 
-// anySeq returns a sequence of objects from a part Iterator.
-func anySeq(iter *part.Iterator[object]) iter.Seq2[any, Revision] {
-	return func(yield func(any, Revision) bool) {
-		// Iterate over a clone of the original iterator to allow the sequence to be iterated
-		// from scratch multiple times.
-		it := iter.Clone()
-		for {
-			_, iobj, ok := it.Next()
-			if !ok {
-				break
-			}
-			if !yield(iobj.data, iobj.revision) {
-				break
-			}
-		}
-	}
-}
-
 // nonUniqueSeq returns a sequence of objects for a non-unique index.
 // Non-unique indexes work by concatenating the secondary key with the
 // primary key and then prefix searching for the items:
 //
-//	<secondary><primary><secondary length>
+//	<secondary>\0<primary><secondary length>
 //	^^^^^^^^^^^
 //
 // Since the primary key can be of any length and we're prefix searching,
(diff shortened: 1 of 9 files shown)


joamaki:
Converted back to draft. There's few more issues with searches into non-unique indexes.

iterator.go
```diff
@@ -132,6 +119,28 @@ func nonUniqueSeq[Obj any](iter *part.Iterator[object], searchKey []byte) iter.S
...
+func nonUniqueLowerBoundSeq[Obj any](iter *part.Iterator[object], searchKey []byte) iter.Seq2[Obj, Revision] {
+	return func(yield func(Obj, Revision) bool) {
+		// Clone the iterator to allow multiple uses.
+		iter = iter.Clone()
+		for {
+			key, iobj, ok := iter.Next()
+			if !ok {
+				break
+			}
+			// With a non-unique index we have a composite key <secondary><primary><secondary len>.
```
bimmlerd, line 131:
```suggestion
			// With a non-unique index we have a composite key <secondary>\0<primary><secondary len>.
```

bimmlerd (approved):
nice find.

txn.go
```diff
@@ -336,30 +338,39 @@ func (txn *txn) delete(meta TableMeta, guardRevision Revision, data any) (object
 	return obj, true, nil
 }
 
+var primaryKeyEncodedLength = base64.StdEncoding.EncodedLen(32)
+
 // encodeNonUniqueKey constructs the internal key to use with non-unique indexes.
 // It concatenates the secondary key with the primary key and the length of the secondary key.
```
bimmlerd, line 344: needs comment on how the primary part is maybe hashed and base64 encoded

bimmlerd (approved):

ovidiutirla (commented):
I mostly have some concerns about the duplicates. As a consumer of those two methods you might probably never want to process the same object twice requiring to add additional logic each time you call Prefix or LowerBound

joamaki:
I redid the encoding schema as base64+sha256 felt like complete overkill. It's now encoding the primary key as follows:
```
  0x0 => 0xfe
  0xfe => 0xfd00
  0xfd => 0xfd01
```

The reasoning is that 0x0 is commonly seen in binary encoded integers and we don't want to expand the length of something like 0x0001, so we substitute that to 0xfe01, and thus we need to substitute 0xfe into something else: 0xfefe => 0xfd00fd00, and now 0xfd needs to become 0xfd01. I chose 0xfe and 0xfd as theyr'e less likely than 0xff for example and not printable chars.

txn.go
joamaki: Redid this using a different schema and documented.

PR #60, opened 2024-10-10
script: Add watch command

joamaki:
The 'watch' command allows watching a table for changes, similar to the `cilium-dbg statedb <table> --watch` flag. The "strike out" support is ported from cilium-dbg/cmd/statedb.go.

The plan is to eventually replace the "cilium-dbg statedb" with the script commands.

```
❯ go run .
> hive start
...
time=2024-10-11T12:25:45.489+02:00 level=INFO msg=Started duration=177.609µs
> db watch ex
Name    X
one     1
two     2
count   0
count   1
count   2
...
^C (interrupted)
> 
```

### DIFF
--- a/script.go
+++ b/script.go
@@ -16,11 +16,12 @@ import (
 	"regexp"
 	"slices"
 	"strings"
-	"text/tabwriter"
 	"time"
 
 	"github.com/cilium/hive"
 	"github.com/cilium/hive/script"
+	"github.com/liggitt/tabwriter"
+	"golang.org/x/time/rate"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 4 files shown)


script.go
```diff
@@ -682,3 +743,95 @@ func joinByPositions(row []string, positions []int) string {
...
+		default:
+			s.buf = append(s.buf, c)
+		}
+	}
+	return len(p), nil
+}
+
+var (
+	// Use color red and the strikethrough escape
+	beginStrikethrough = []byte("\033[9m\033[31m")
```
joamaki, line 821: Of course slightly naughty doing ansi escapes here. I'm leaning towards just YOLO'ing on this as it's for just the watch command and you want to run it interactively. Alternatively could mark differently, e.g. with `<<deleted>>` and translate to ansi codes on the `cilium-dbg` etc. side.

pippolo84: Agree with doing this here, since `watch` is interactive. We can rework it later if needed.

pippolo84 (approved):
:100:

PR #62, opened 2024-10-29
Fix JSON formatting in db show command

joamaki:
The "db show -format=json" was missing newline at the end. While at it, switch to MarshalIndent for nicer output and extend the tests.

Hive dependency is bumped to add the -scripttest.update support.


bimmlerd (approved):
nice, love the tests

PR #64, opened 2024-11-04
Simplify tracking of deleted objects

joamaki:
This reworks how deleted objects are stored by removing the separate graveyard indexes and instead storing the deleted objects in the normal primary and revision indexes. This simplifies the code fair bit, especially for Changes(), and allows for more efficient cleanup of deleted objects as part of Commit() rather than in a separate graveyard worker.

```
benchmarks: Add benchmark for deletion

    Add a benchmark for Delete() to test impact of the graveyard index removal.

Reimplement the deleted object tracking

    The use of separate graveyard primary and revision indexes and a background
    GC job to remove observed objects from them was unnecessarily complicated.

    We can implement this in a much simpler way by keeping the
    "soft-deleted" objects in the normal primary & revision indexes, but just
    marking them as deleted.

    The garbage collection of observed dead objects can be performed as part of
    the Commit(), which has the benefit that we can operate on already cloned
    mutated nodes, allowing in-place modification of the trees. This way we also
    don't need any background jobs in StateDB. This does have the semantic
    difference that observed deleted objects are only dropped from indexes on a
    subsequent WriteTxn, but I don't we'd have an issue with the delayed
    deletions in practice. If needed we can add back a background job to
    essentially do a 'WriteTxn(allTables...).Commit()' periodically.

    There's not a huge difference in benchmarks in terms of time since we're 
    essentially saving a lookup of the graveyard tree during inserts, but this 
    does allow for more compact indexing due to fewer and more optimal radix 
    tree nodes due to the merged indexes. The code for iterating over changes is
    also a bit more efficient and simpler as it's now just an iteration over a
    single index.
```


metrics.go
```diff
@@ -75,15 +70,14 @@ func NewExpVarMetrics(publish bool) *ExpVarMetrics {
...
-		GraveyardCleaningDurationVar: newMap("graveyard_cleaning_duration"),
-		GraveyardLowWatermarkVar:     newMap("graveyard_low_watermark"),
-		GraveyardObjectCountVar:      newMap("graveyard_object_count"),
-		ObjectCountVar:               newMap("object_count"),
-		WriteTxnAcquisitionVar:       newMap("write_txn_acquisition"),
-		WriteTxnDurationVar:          newMap("write_txn_duration"),
-		DeleteTrackerCountVar:        newMap("delete_tracker_count"),
-		RevisionVar:                  newMap("revision"),
+		LockContentionVar:        newMap("lock_contention"),
+		GraveyardLowWatermarkVar: newMap("graveyard_low_watermark"),
```
joamaki, line 74: I think keeping the graveyard naming makes sense here as `deleted_object_count` would be rather confusing.

txn.go
```diff
@@ -303,40 +299,46 @@ func (txn *txn) delete(meta TableMeta, guardRevision Revision, data any) (object
...
 	// We assume that "data" has only enough defined fields to
 	// compute the primary key.
 	idKey := meta.primary().fromObject(object{data: data}).First()
 	idIndexTree := txn.mustIndexWriteTxn(meta, PrimaryIndexPos)
-	obj, existed := idIndexTree.Delete(idKey)
-	if !existed {
+	var obj object
+	var existed bool
+	if txn.hasDeleteTrackers(meta) {
+		obj, _, existed = idIndexTree.Get(idKey)
```
joamaki, line 315: I originally wrote this using `Modify` and also changed part.Modify to support not inserting, but it was still cloning the nodes on the way and it was a bit complex. I didn't see any performance difference between Get+Insert and the Modify so ended up just throwing those changes away.

iterator.go
```diff
@@ -134,6 +138,10 @@ func nonUniqueSeq[Obj any](iter *part.Iterator[object], prefixSearch bool, searc
 				visited[string(primary)] = struct{}{}
 			}
 
+			if iobj.deleted {
```
bimmlerd, line 141: could do this before even decoding the key?

iterator.go
```diff
@@ -172,93 +185,20 @@ func nonUniqueLowerBoundSeq[Obj any](iter *part.Iterator[object], searchKey []by
...
 type changeIterator[Obj any] struct {
-	table          Table[Obj]
-	revision       Revision
-	deleteRevision Revision
-	dt             *deleteTracker[Obj]
-	iter           *DualIterator[Obj]
-	watch          <-chan struct{}
+	table         Table[Obj]
+	revision      Revision
+	startRevision Revision
```
bimmlerd, line 198: given we have a revision and a startRevision maybe comment on what's what

iterator.go
```diff
@@ -269,9 +209,7 @@ func (it *changeIterator[Obj]) refresh(txn ReadTxn) {
 	itxn := txn.getTxn()
 	indexEntry := itxn.root[it.table.tablePos()].indexes[RevisionIndexPos]
 	indexTxn := indexReadTxn{indexEntry.tree, indexEntry.unique}
-	updateIter := &iterator[Obj]{indexTxn.LowerBound(index.Uint64(it.revision + 1))}
-	deleteIter := it.dt.deleted(itxn, it.deleteRevision+1)
-	it.iter = NewDualIterator(deleteIter, updateIter)
+	it.iter = indexTxn.LowerBound(index.Uint64(it.revision + 1))
 
 	// It is enough to watch the revision index and not the graveyard since
```
bimmlerd, line 214: no more graveyard, so comment can go
generally probably grep for graveyard; there's some more comment references to it which are stale

iterator.go
```diff
@@ -307,17 +245,24 @@ func (it *changeIterator[Obj]) Next(txn ReadTxn) (seq iter.Seq2[Change[Obj], Rev
...
+			it.revision = rev
 			change := Change[Obj]{
-				Object:   obj,
 				Revision: rev,
-				Deleted:  deleted,
+			}
+			if obj.deleted {
+				if rev <= it.startRevision {
+					// Ignore objects that were marked deleted before this
+					// changge iterator was created.
```
bimmlerd, line 257: changge typo

It's somehow asymmetric to me that the iterator has a startRevision, but only cares about it for deleted objects. It seems correct, but irks me for some reason

metrics.go
bimmlerd: I guess it would be `objects_pending_deletion_count`, but yeah, graveyard here is more concise.

table.go
```diff
@@ -388,7 +391,7 @@ func (t *genTable[Obj]) ListWatch(txn ReadTxn, q Query[Obj]) (iter.Seq2[Obj, Rev
 		// Doing a Get() is more efficient than constructing an iterator.
 		value, watch, ok := indexTxn.Get(q.key)
 		seq := func(yield func(Obj, Revision) bool) {
-			if ok {
+			if ok && !value.deleted {
 				yield(value.data.(Obj), value.revision)
 			}
 		}
```
bimmlerd, line 397: Also in the CompareAndSwap impl; should the compare not first check that the compared to obj is not deleted?

txn.go
```diff
@@ -621,3 +646,46 @@ func (txn *txn) WriteJSON(w io.Writer, tables ...string) error {
...
+			_, hadOld := revTree.Delete(revKey[:])
+			if !hadOld {
+				panic("BUG: Object to be deleted not found from revision index")
+			}
+			idKey := primIndexer.fromObject(object{data: obj.data}).First()
+			_, hadOld = primTree.Delete(idKey)
+			if !hadOld {
+				panic("BUG: Object to be deleted not found from primary index")
+			}
+			table.numDeletedObjects--
```
bimmlerd, line 683: I guess you could call it `numObjsPendingDeletion` instead, which would look more intuitive here, but it's fine either way

bimmlerd (approved):
couple of nits, but haven't spotted bugs so far. it seems pretty easy to miss a `if obj.deleted` somewhere, but I'd hope tests catch it

types.go
```diff
@@ -387,23 +387,18 @@ type TableWritable interface {
...
+	RevisionIndex          = "__revision__"
+	RevisionIndexPos       = 1
+	SecondaryIndexStartPos = 2
 )
 
 // object is the format in which data is stored in the tables.
 type object struct {
 	revision uint64
 	data     any
+	deleted  bool
```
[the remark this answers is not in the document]
joamaki: I played a bit with having `data` point to `type struct deleted { data any }`, but that just caused a further allocation so was no good. If we mandated that `T` is always a pointer then we could store `data unsafe.Pointer` here and use the bytes saved from the not having the type information around, but that's probably a bit annoying to work with. 

But yeah, probably worth checking the field alignments, especially with the object embedded into a part node. Will take a look.

types.go
```diff
@@ -440,28 +435,44 @@ type indexEntry struct {
...
+
 	revision            uint64
 	pendingInitializers []string
 	initialized         bool
 	initWatchChan       chan struct{}
+
+	// deletedRange holds the revisions of the oldest and
+	// newest deleted objects. Used to short-cut GCing of
+	// deleted objects.
+	deletedRange revisionRange
```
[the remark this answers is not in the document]
joamaki: I like `gcRange`

types.go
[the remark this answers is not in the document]
bimmlerd: could also halve the revision space and use the bit 😇 
but I think you mentioned somewhere that the size didn't increase? that's surprising 😅

table.go
joamaki: Not sure what you mean. We can't look into `value` unless `ok` is true.

iterator.go
joamaki: Yeah let me rename that!

types.go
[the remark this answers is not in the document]
joamaki: Hmm, we could do that, but need to make sure the "deleted bit" is the least significant one which is a bit annoying (shift left by one and set). But yeah since the size didn't increase (most likely because there were padding in the node structs) not worth optimizing.

table.go
bimmlerd: uhh, I think I was trying to comment on CompareAndDelete below but couldn't because GitHub UI

but the inner `delete` handles the `deleted` objects so I think this is just fine

joamaki:
While I would've guessed that this approach would've performed better and used less memory the statistics from the load-balancer benchmark show otherwise:

With this change:
```
Memory statistics from N=10 iterations:
Min: Allocated 815937kB in total, 3162627 objects / 208585kB still reachable (per service:  63 objs, 16710B alloc,  4271B in-use)
Avg: Allocated 825660kB in total, 3183026 objects / 216537kB still reachable (per service:  63 objs, 16909B alloc,  4434B in-use)
Max: Allocated 900403kB in total, 3332232 objects / 277196kB still reachable (per service:  66 objs, 18440B alloc,  5676B in-use)
```

Before this:
```
Memory statistics from N=10 iterations:
Min: Allocated 818803kB in total, 2294293 objects / 127964kB still reachable (per service:  45 objs, 16769B alloc,  2620B in-use)
Avg: Allocated 833869kB in total, 2586240 objects / 163081kB still reachable (per service:  51 objs, 17077B alloc,  3339B in-use)
Max: Allocated 902984kB in total, 3329752 objects / 276950kB still reachable (per service:  66 objs, 18493B alloc,  5671B in-use)
```

Because of that I won't move forward this at least in this form and will close this PR for now.

PR #65, opened 2024-11-07
Improve db command help

joamaki:
Add a long-form help text for the db command that is shown with 'help db'.

### DIFF
--- a/script.go
+++ b/script.go
@@ -25,6 +25,66 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+func underline(s string) string {
+	return "\033[4m" + s + "\033[0m"
+}
+
+func usageDetails(cmds map[string]script.Cmd, sortedNames []string) (out []string) {
+	out = strings.Split(`DESCRIPTION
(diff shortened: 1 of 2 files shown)


script.go
```diff
@@ -160,8 +238,21 @@ func InitializedCmd(db *DB) script.Cmd {
...
 		script.CmdUsage{
-			Summary: "Show table",
-			Args:    "(-o=<file>) (-columns=col1,...) (-format={table,yaml,json}) table",
+			Summary: "Show the contents of a table",
+			Args:    "[-o=<file>] [-columns=col1,...] [-format={table,yaml,json}] table",
+			Detail: []string{
+				"Show the contents of a table.",
+				"",
+				"The contents are written to stdout, but can be written to",
+				"a file instead by with the -o flag.",
```
ovidiutirla, line 247:
```suggestion
				"a file instead with the -o flag.",
```

ovidiutirla (approved):
🚀

script.go
joamaki: Thanks, fixed!

PR #66, opened 2024-11-07
reconciler: Fix flake in reconciler tests due to ID being unset

joamaki:
The fix to JSON status marshalling had the unintended effect that now the reconciler test suite ran with the ID field being uninitialized, leading to the reconciler overwriting the status of a changed object. This resulted in small fraction of the reconciler test runs failing (e.g. object was marked pending, but instead of being re-reconciled the previous status was inserted).

This did not affect non-test code as status should always be constructed with StatusPending() etc.

Fix this by exporting the ID and filling it in the tests.

Fixes: 132edd67f709 ("reconciler: Fix Status JSON marshalling")

### DIFF
--- a/reconciler/incremental.go
+++ b/reconciler/incremental.go
@@ -174,7 +174,7 @@ func (round *incrementalRound[Obj]) batch(changes iter.Seq2[statedb.Change[Obj],
 			if entry.Result == nil {
 				round.retries.Clear(entry.Object)
 			}
-			round.results[entry.Object] = opResult{rev: entry.Revision, id: status.id, err: entry.Result, original: entry.original}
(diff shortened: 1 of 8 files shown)


joamaki:
Failure can be seen in https://github.com/cilium/statedb/actions/runs/11707677739/job/32607608514:
```
            > db show test-objects
            [stdout]
            ID   Faulty   StatusKind   StatusError
            1    false    Pending      
            2    false    Done         
            3    false    Done         
            > db cmp test-objects step1+3.table
        scripttest.go:230: FAIL: testdata/incremental.txtar:27: db cmp test-objects step1+3.table: table mismatch:
              ID  StatusKind StatusError
            - 1   Error      update fail
            + 1   Done
              2   Done       
              3   Done       
```

bimmlerd (approved):
ah, right. Ouch.

PR #67, opened 2024-11-07
part: Fix MarshalYAML on empty maps

joamaki:
MarshalYAML nil deref'd the 'm.tree' when the map was empty. Fix this and extend test coverage.

### DIFF
--- a/part/map.go
+++ b/part/map.go
@@ -243,9 +243,11 @@ func (m *Map[K, V]) UnmarshalJSON(data []byte) error {
 
 func (m Map[K, V]) MarshalYAML() (any, error) {
 	kvs := make([]mapKVPair[K, V], 0, m.Len())
-	iter := m.tree.Iterator()
-	for _, kv, ok := iter.Next(); ok; _, kv, ok = iter.Next() {
-		kvs = append(kvs, kv)
+	if m.tree != nil {
+		iter := m.tree.Iterator()
+		for _, kv, ok := iter.Next(); ok; _, kv, ok = iter.Next() {
+			kvs = append(kvs, kv)
+		}
 	}
 	return kvs, nil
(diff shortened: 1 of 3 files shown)


joamaki:
Test failing due to https://github.com/cilium/statedb/pull/66. And it looks like I've made the test pass even when it fails in the PR that changed the workflow to post the message even when it fails. Oops. Need to look into that..

part/map.go
```diff
@@ -243,9 +243,11 @@ func (m *Map[K, V]) UnmarshalJSON(data []byte) error {
...
 func (m Map[K, V]) MarshalYAML() (any, error) {
 	kvs := make([]mapKVPair[K, V], 0, m.Len())
-	iter := m.tree.Iterator()
-	for _, kv, ok := iter.Next(); ok; _, kv, ok = iter.Next() {
-		kvs = append(kvs, kv)
+	if m.tree != nil {
+		iter := m.tree.Iterator()
+		for _, kv, ok := iter.Next(); ok; _, kv, ok = iter.Next() {
+			kvs = append(kvs, kv)
+		}
```
ovidiutirla, line 250:
```suggestion
	if m.tree == nil {
		return kvs, nil
	}

	iter := m.tree.Iterator()
	for _, kv, ok := iter.Next(); ok; _, kv, ok = iter.Next() {
		kvs = append(kvs, kv)
```
WDYT about flipping the if statement?

ovidiutirla (approved):
🚀

part/map.go
joamaki: I think in this case I prefer the way I wrote it: you start with empty and you populate it when there's values to put in. it's much shorter too. Though in general it's often better to do the early return.

PR #68, opened 2024-11-07
CI: Fail action if make fails (pipefail oops)

joamaki:
The posting of the test run as a comment used `make | tee out` which without `pipefail` ate the exit status from `make`. Use `set -o pipefail` in the jobs to properly propagate this. While at it, simplify the workflow a bit.

In case you're wondering why post a comment at all: this is my simple way of easily keeping track of the code test coverage and the benchmark numbers (I always check the reconciler throughput for example). This will also leave a trace of which PR potentially started causing regressions (unreliable of course due to benchmarks running potentially different hardware, but it's something).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,26 +17,21 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+
       - name: test
         run: |
+          set -o pipefail
           echo '```' > results.comment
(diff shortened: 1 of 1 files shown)


joamaki:
Example run where this catches the failure properly: https://github.com/cilium/statedb/actions/runs/15187170836/job/42710304722?pr=68

joestringer (approved):

PR #69, opened 2024-11-11
Flatten the 'db' commands, e.g. 'db prefix' to 'db/prefix'

joamaki:
The user experience with commands that have sub-commands is not great and it makes it harder to implement completion in 'cilium-dbg shell' if we so choose. There isn't really any good rational for nesting the commands like this, so let's go with Plan 9 style command names (https://9p.io/magic/man2html/8/auth), e.g. "db prefix" becomes "db/prefix".

### DIFF
--- a/reconciler/testdata/batching.txtar
+++ b/reconciler/testdata/batching.txtar
@@ -7,10 +7,10 @@ start-reconciler with-batchops
 # From here this is the same as incremental.txtar.
 
 # Step 1: Insert non-faulty objects
-db insert test-objects obj1.yaml
-db insert test-objects obj2.yaml
-db insert test-objects obj3.yaml
-db cmp test-objects step1+3.table
+db/insert test-objects obj1.yaml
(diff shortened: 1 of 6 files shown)


joamaki:
I went to fair bit of effort in https://github.com/cilium/statedb/pull/65 to document the "sub-commands" and I just realized today that it's kind of stupid to nest the commands like this. It makes them hard to find (`help` wont list them) and it makes it hard/impossible to implement command-line completion in `cilium-dbg shell` if we want to add one later. The script/shell thing should be as simple and dumb as possible, so let's simplify this.

ovidiutirla (approved):
🚀

pippolo84 (approved):
Nice to see that the (pretty complicated) logic to generate the usage details has gone away. I'm (mildly) concerned that the Plan 9 style might be kind of unexpected when using statedb in cilium where we rely on subcommands, but I'd say that it's worth to give this a try. :+1:

PR #70, opened 2024-12-13
Refactor to use the new flags system in script

joamaki:
Use the new pflag-based system for specifying flags for script commands.


pippolo84 (approved):
:100:

PR #72, opened 2025-01-28
Add WatchSet for watching a dynamic set of channels

joamaki:
When writing a controller that needs to react to specific queries changing it's useful to be able to watch a dynamic set of channels. Add the [WatchSet] utility for this.

Example use:
```
  var names []string = ...
  ws := statedb.NewWatchSet()
  for {
    for _, name := range names {
      things, watch := myTable.ListWatch(txn, ThingsByName(name))
      ws.Add(watch)
      processThings(things)
    }
    // Wait for things that have a name in 'names' to change.
    if err := ws.Wait(ctx); err != nil {
      // Context cancelled
      break
    }
  }
```

### DIFF
--- a/watchset.go
+++ b/watchset.go
@@ -1,0 +1,103 @@
+package statedb
+
+import (
+	"context"
+	"slices"
+	"sync"
+)
+
+const watchSetChunkSize = 16
+
+// WatchSet is a set of watch channels that can be waited on.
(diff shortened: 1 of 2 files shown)


pippolo84 (approved):
Love the idea of chunking and unrolling to boost efficiency, nice! :100:

PR #74, opened 2025-02-10
watchset: Use a map for channels and add Merge and Has

joamaki:
It is not very uncommon that we'd add duplicate channels to the watch set, so better deduplicate.

Add Merge() and Has() methods. These are useful when we need to know whether a closed channel affected a particular subset of work.


pippolo84 (approved):
LGTM

PR #77, opened 2025-03-24
Avoid repeated Clone() when doing reads with a WriteTxn

joamaki:
This reduces allocations a bit when using `WriteTxn` to do `Prefix` or `LowerBound` queries by not repeatedly cloning the `part.Txn` of a `WriteTxn` when no writes have been made.

Benchmark before:
```
    BenchmarkDB_SequentialInsert_Prefix-32               207           5982505 ns/op            167154 objects/sec   5611786 B/op      67561 allocs/op
```

after:
```
    BenchmarkDB_SequentialInsert_Prefix-32               247           4594889 ns/op            217633 objects/sec   3795819 B/op      58560 allocs/op
```

The benchmark does an Insert followed by 5 prefix searches for the inserted object. ~15% less allocations and ~20% faster benchmark result.

### DIFF
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -231,6 +231,31 @@ func BenchmarkDB_SequentialInsert(b *testing.B) {
 	b.ReportMetric(float64(numObjectsToInsert*b.N)/b.Elapsed().Seconds(), "objects/sec")
 }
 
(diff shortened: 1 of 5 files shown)


dylandreimerink (approved):
:broom:

PR #81, opened 2025-06-18
part: RootOnlyWatch fixes

joamaki:
The InsertWatch/ModifyWatch returned nil channel with RootOnlyWatch. Should always return the root channel.

The root node watch channel was not replaced on promotion with RootOnlyWatch. This potentially impacts StateDB as it uses RootOnlyWatch for the revision index so when revision `2^58` (6th most significant bit changes) is hit the root node will need to split from node4 to node16 hitting the bug.

### DIFF
--- a/part/txn.go
+++ b/part/txn.go
@@ -217,7 +217,7 @@ func (txn *Txn[T]) modify(root *header[T], key []byte, mod func(T) T) (oldValue 
 				if this.watch != nil {
 					txn.watches[this.watch] = struct{}{}
 				}
-				this = this.promote(!txn.opts.rootOnlyWatch || this == newRoot)
+				this = this.promote(!txn.opts.rootOnlyWatch || this == root)
 				txn.mutated.put(this)
 			} else {
 				// Node is big enough, clone it so we can mutate it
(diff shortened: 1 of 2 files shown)


part/txn.go
```diff
@@ -217,7 +217,7 @@ func (txn *Txn[T]) modify(root *header[T], key []byte, mod func(T) T) (oldValue
 				if this.watch != nil {
 					txn.watches[this.watch] = struct{}{}
 				}
-				this = this.promote(!txn.opts.rootOnlyWatch || this == newRoot)
+				this = this.promote(!txn.opts.rootOnlyWatch || this == root)
```
bimmlerd, line 220: not of consequence as far as I can tell since it's written to before read, but `thisp` in line 198 should probably point to `&root` instead of `&newRoot`? Or just nil, if it really is always written to before being read

bimmlerd (approved):

part/txn.go
joamaki: No it must point to `&newRoot` so that we'll update it correctly as we recurse into the old tree and clone it.

bimmlerd: oh duh, yeah I get it now :+1:

PR #85, opened 2025-06-19
part: Optimize map space usage with one element

joamaki:
In cilium/cilium we're using part.Map in the Backend object to hold the instances of the backend and most of the time it only contains one element.

Specialize for this case by adding a singleton field to part.Map.

Before (map with 1 element):
```
=== RUN   TestMapMemoryUse
map_test.go:272: 320 bytes per map
```

After:
```
=== RUN   TestMapMemoryUse
map_test.go:348: 40 bytes per map
```

pkg/loadbalancer/benchmark before:
```
Memory statistics from N=10 iterations:
Min: Allocated 716852kB in total, 2068061 objects / 125960kB still reachable (per service:  41 objs, 14681B alloc,  2579B in-use) 
Avg: Allocated 731848kB in total, 2607405 objects / 180868kB still reachable (per service:  52 objs, 14988B alloc,  3704B in-use) 
Max: Allocated 775441kB in total, 3649005 objects / 317829kB still reachable (per service:  72 objs, 15881B alloc,  6509B in-use)

Insert statistics from N=10 iterations:
Min: Reconciled 50000 objects in 832.079251ms (16.641µs  per service /  60093 services per second) 
Avg: Reconciled 50000 objects in 883.983229ms (17.679µs  per service /  56564 services per second) 
Max: Reconciled 50000 objects in 988.247623ms (19.764µs  per service /  50597 services per second)
```

After:
```
Memory statistics from N=10 iterations:
Min: Allocated 669824kB in total, 2067737 objects / 124777kB still reachable (per service:  41 objs, 13718B alloc,  2555B in-use) 
Avg: Allocated 688314kB in total, 2445133 objects / 166797kB still reachable (per service:  48 objs, 14096B alloc,  3416B in-use) 
Max: Allocated 734497kB in total, 3444117 objects / 301701kB still reachable (per service:  68 objs, 15042B alloc,  6178B in-use)

Insert statistics from N=10 iterations:
Min: Reconciled 50000 objects in 798.811656ms (15.976µs  per service /  62594 services per second) 
Avg: Reconciled 50000 objects in 831.351251ms (16.627µs  per service /  60143 services per second) 
Max: Reconciled 50000 objects in 929.973547ms (18.599µs  per service /  53766 services per second)
```

### DIFF
--- a/part/map.go
+++ b/part/map.go
@@ -19,8 +19,9 @@ import (
 // Map is a typed wrapper around Tree[T] for working with
 // keys that are not []byte.
 type Map[K, V any] struct {
-	bytesFromKey func(K) []byte
-	tree         *Tree[mapKVPair[K, V]]
+	bytesFromKeyFunc func(K) []byte
+	tree             *Tree[mapKVPair[K, V]]
+	singleton        *mapKVPair[K, V]
 }
 
 type mapKVPair[K, V any] struct {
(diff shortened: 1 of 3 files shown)


part/map.go
```diff
@@ -167,6 +226,9 @@ func (m Map[K, V]) SlowEqual(other Map[K, V]) bool {
 		return true
```
bimmlerd, line 226: I think this case is broken for two singletons with differing contents. Needs the same logic as above. (And possibly a test which tests SlowEqual for this case)

bimmlerd (changes requested):

part/map.go
joamaki: Oh indeed, forgot to update that. Fixed and added a quick test to validate the properties.

PR #86, opened 2025-06-20
part: Clear tree.txn after reusing it

joamaki:
To allow forking the tree into two lineages we'll need to make sure only the first Txn() call will reuse the allocation and pass it to the produced tree.

Also drop the pre-allocation of 'Txn' in 'New'. It'll be produced by the first 'Txn()' call and we'll use less memory for empty trees this way.

### DIFF
--- a/part/part_test.go
+++ b/part/part_test.go
@@ -1003,6 +1003,25 @@ func Test_lowerbound_bigger(t *testing.T) {
 	require.False(t, ok)
 }
 
+// Test that we can "fork" the tree even though we're reusing the txn allocation.
+func Test_fork_with_cache(t *testing.T) {
+	tree := New[uint64]()
+	_, _, tree = tree.Insert(uint64Key(1), 1)
+
+	txn1 := tree.Txn()
+	txn2 := tree.Txn()
+	txn1.Insert(uint64Key(2), 2)
+	txn2.Insert(uint64Key(3), 3)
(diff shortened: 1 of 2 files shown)


part/part_test.go
```diff
@@ -1003,6 +1003,25 @@ func Test_lowerbound_bigger(t *testing.T) {
 	require.False(t, ok)
 }
 
+// Test that we can "fork" the tree even though we're reusing the txn allocation.
+func Test_fork_with_cache(t *testing.T) {
+	tree := New[uint64]()
+	_, _, tree = tree.Insert(uint64Key(1), 1)
+
+	txn1 := tree.Txn()
+	txn2 := tree.Txn()
```
bimmlerd, line 1012: I'm confused - I thought that only a single in-flight Txn against a part tree is valid.

joamaki: Yes if watch channels are used, e.g. `Commit()` is called as otherwise we'd get a "close of a closed channel", but it's not true if we're not using watch channels and only calling `CommitOnly()`. This is the case for `part.Map` and `part.Set`.

bimmlerd: Ah - right, I see. Probably add this wording to the comment above `Txn()` - that's where I had the single in-flight thing from.

Yeah the mutable marker might be a nice thing to catch errors

PR #89, opened 2025-06-26
part: Fix singleton case in Map.Delete

joamaki:
Delete was missing the conversion into a singleton on deletion. Extend the map quick tests to check the singleton properties.


bimmlerd (approved):
LGTM

PR #90, opened 2025-06-26
Fixes to Prefix() and GetWatch()

joamaki:
GetWatch() returned a mismatching leaf watch channel if there was a leaf with a key shorter than search key.
Prefix() returned root watch channel instead of the closest non-leaf channel.


bimmlerd (approved):
yep, both fixes and tests seem sound

PR #94, opened 2025-07-25
statedb: improve handling of initializers

giorio94:
Currently, [Table.Initialized] is affected by a potential pitfall, as it can return false (i.e., not initialized) and a closed watch channel if any write transaction (e.g., to populate an initial state, or register a changes iterator) completed before the first initializer is registered. In turn, this can cause consumers to misbehave, e.g., entering a busy loop, or incorrectly thinking that the table is initialized, if relying on the channel only. 

This commit addresses this issue by registering an implicit initializer that gets automatically lifted once the DB starts. In other words, any table, regardless of the presence of custom initializers, is not treated as initialized until the DB starts.

This approach ensures that initializers can be safely registered before the DB start operation, guaranteeing that they can always take effect regardless of the ordering of operations (e.g., with respect to checking for initialization, or the execution of parallel write transactions).

The registration of initializers after that the DB has started is now also explicitly forbidden (i.e., triggers a panic), given that susceptible to misuse.

Additionally, let's also panic if initializer is marked done while the target table is not included in the write transaction, to make the failure explicit, instead of silently skipping the operation and not marking the table as initialized.

### DIFF
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -26,7 +26,7 @@ import (
 const numObjectsToInsert = 1000
 
 func BenchmarkDB_WriteTxn_1(b *testing.B) {
-	db, table := newTestDBWithMetrics(b, &NopMetrics{})
+	db, table := newTestDB(b)
 
 	for b.Loop() {
 		txn := db.WriteTxn(table)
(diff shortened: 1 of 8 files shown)


giorio94:
Closing in favor of https://github.com/cilium/statedb/pull/95, which implements the alternative approach described in the comment above.

PR #100, opened 2025-07-31
statedb: Add TableWritable as constraint on NewTable, add NewTableAny

joamaki:
Let's start enforcing the TableWritable on the tables so we have proper "db/show" output for all tables.

To make it possible to have StateDB tables without having to introduce a wrapper type for types we don't control add a NewTableAny constructor which takes TableHeader() and TableRow() as arguments.


giorio94 (approved):
Thanks!

PR #101, opened 2025-07-31
reconciler: Shrink Status type

joamaki:
1. Change `Error` from `string` to `*string` as it's rarely set (64 => 56 bytes)
2. Change `StatusKind` from `string` to `uint8` (56 => 48 bytes)

The `StatusKind` is `struct { uint8 }` to make sure code that used to do `string(foo.Status.Kind)` won't compile and is forced to change to `foo.Status.Kind.String()`. Note that we're going to release v0.5.0 with few other breaking API changes in addition to this one and tooling in cilium/cilium only bumps StateDB on patch version changes.

### DIFF
--- a/reconciler/types.go
+++ b/reconciler/types.go
@@ -133,7 +133,7 @@ func (s StatusKind) Key() index.Key {
 type Status struct {
 	Kind      StatusKind `json:"kind" yaml:"kind"`
 	UpdatedAt time.Time  `json:"updated-at" yaml:"updated-at"`
-	Error     string     `json:"error,omitempty" yaml:"error,omitempty"`
+	Error     *string    `json:"error,omitempty" yaml:"error,omitempty"`
 
 	// id is a unique identifier for a pending object.
 	// The reconciler uses this to compare whether the object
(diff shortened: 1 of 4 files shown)


joamaki:
On `main` the reconciler benchmark gives:
```
1166MB total allocated, 6011116 in-use objects, 353MB bytes in use
```

And in this PR:
```
1137MB total allocated, 6011120 in-use objects, 338MB bytes in use
```

So for 1M objects with `Status` we're using ~4% less memory.

reconciler/types.go
```diff
@@ -196,18 +256,19 @@ func StatusDone() Status {
...
-		Error:     "",
+		Error:     nil,
 		ID:        nextID(),
 	}
 }
 
 // statusError constructs the status that marks the object
 // as failed to be reconciled.
 func StatusError(err error) Status {
+	errStr := err.Error()
```
derailed, line 267: this will panic if called with nil

reconciler/types.go
```diff
@@ -94,15 +96,65 @@ type BatchOperations[Obj any] interface {
...
+type StatusKind struct {
+	kind uint8
+}
 
-const (
-	StatusKindPending    StatusKind = "Pending"
-	StatusKindRefreshing StatusKind = "Refreshing"
-	StatusKindDone       StatusKind = "Done"
-	StatusKindError      StatusKind = "Error"
+var (
```
derailed, line 108: Could be out of context here but this feels a little odd aka consts -> global variables. Not sure if applicable but perhaps we could have gone with StatusKind uint8 and generate the string equivalent of the consts via go:generate. 
Aka have the compiler generate the map vs manually curating `statusKindToString`

derailed (approved):
@joamaki Nice work! few comments but out of context so feel free to discard...

reconciler/types.go
joamaki: I wish I could do that, but since we have existing code doing `string(foo.Status.Kind)` we cannot do `type StatusKind uint8` since `string(uint8(0))` compiles and we want to force the code to switch to `foo.Status.Kind.String()`. Maybe in the future once all use-sites have converted over? For now I'd like to err on the safe side.

PR #103, opened 2025-08-28
statedb: Clarify primary index key requirements

gandro:
The primary index seems to have the requirement that every object must have not less and not more than one key.


pippolo84 (approved):
Thanks!

I guess we should follow up enforcing this with a panic. Might be worth open an issue?

gandro:
Something that came to my mind later: Does this restriction also apply to non-primary unique indices? My current superficial understanding is that this restriction exists mainly on the primary index, but not for other unique indices.

PR #104, opened 2025-09-04
reconciler: Take a job.Group instead of creating one

joamaki:
To allow using reconciler.Register after the application has started change it to take a job.Group to which to append its jobs instead of having it create one.

### DIFF
--- a/reconciler/script_test.go
+++ b/reconciler/script_test.go
@@ -34,13 +34,12 @@ func newEngine(t testing.TB, args []string) *script.Engine {
 	log := hivetest.Logger(t)
 
 	var (
-		ops                 = &mockOps{}
-		db                  *statedb.DB
-		r                   reconciler.Reconciler[*testObject]
-		reconcilerParams    reconciler.Params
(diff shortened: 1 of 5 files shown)


reconciler/script_test.go
```diff
@@ -65,22 +64,18 @@ func newEngine(t testing.TB, args []string) *script.Engine {
...
 				func() reconciler.Metrics {
 					return expVarMetrics
-				}),
+				},
+				job.Registry.NewGroup,
+			),
 
 			cell.Invoke(
-				func(db_ *statedb.DB, p_ reconciler.Params) (err error) {
+				func(db_ *statedb.DB, p_ reconciler.Params, lc cell.Lifecycle) (err error) {
```
pippolo84, line 72: `lc` is actually unused, we can get rid of it.

pippolo84 (approved):
LGTM, just a minor thing left inline

gandro (approved):
Not familiar enough with Hive to judge correctness, but thank you for the change!

PR #106, opened 2025-10-01
script: Allow dynamic comparison in "db/cmp" using env vars

christarazi:
- **script: Expand env inside db/cmp command**
- **script: Allow dynamic comparison when env vars are used**

### DIFF
--- a/script.go
+++ b/script.go
@@ -351,7 +351,7 @@ func CompareCmd(db *DB) script.Cmd {
 			if err != nil {
 				return nil, fmt.Errorf("ReadFile(%s): %w", args[1], err)
 			}
-			lines := strings.Split(string(data), "\n")
+			lines := strings.Split(s.ExpandEnv(string(data), false), "\n")
 			lines = slices.DeleteFunc(lines, func(line string) bool {
 				return strings.TrimSpace(line) == ""
 			})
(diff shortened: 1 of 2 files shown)


script.go
```diff
@@ -934,19 +935,29 @@ func splitHeaderLine(line string) (names []string, pos []int) {
 // The whitespace on the right of the start position (e.g. "1  \t") is trimmed.
 // This of course requires that the table is properly formatted in a way that the
 // header columns are indented to fit the data exactly.
-func splitByPositions(line string, positions []int) []string {
+func splitByPositions(line string, positions []int, splitByTabs bool) []string {
```
pippolo84, line 938: Question: did you try to add a separate `splitByPositionsWithTabs` function? I wonder if a bit of repetition in place of the `if` might help readability. Same for `joinByPositions.

pippolo84 (commented):
I'm not sure I got the relation between the tabs based indentation and the env var expansion. Adding a test in `db.txtar` for this scenario should help clarify that.

script.go
christarazi: I've updated the commit msg to clarify further and added an example. I considered refactoring but thought maybe it's easier to keep it concise. I'm happy to change it if it is still not very readable after the updated commit msg and example.

pippolo84: No need to do that, it was just an idea. We can still rework this later if further changes to the way we split are needed :+1:

pippolo84 (approved):
:heavy_check_mark:

PR #107, opened 2025-10-02
part: Improve node4 find speed by ~2%

joamaki:
Change the loop into a switch. We don't need to check the size as all operations make sure to nil out the child.

Before:
BenchmarkDB_RandomLookup-6                         27901             42534 ns/op          23510630 objects/sec
BenchmarkDB_SequentialLookup-6                     35810             33500 ns/op          29851094 objects/sec

After:
BenchmarkDB_RandomLookup-6                         28411             41583 ns/op          24048128 objects/sec
BenchmarkDB_SequentialLookup-6                     36318             32940 ns/op          30358678 objects/sec

41583/42534 = 0.977
32940/32940 = 0.98i3

### DIFF
--- a/part/node.go
+++ b/part/node.go
@@ -332,13 +332,16 @@ func (n *header[T]) find(key byte) *header[T] {
 		return nil
 	case nodeKind4:
 		n4 := n.node4()
-		size := n4.size()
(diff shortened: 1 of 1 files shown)


derailed (approved):
@joamaki Nice!

PR #158, opened 2026-05-05
reconciler: append name to job names

giorio94:
Respect the reconciler name configured through the [WithName] option when constructing the name of the jobs, to prevent them from conflicting (and consequently have the health scopes conflict) in case of multiple reconcilers inside the same module.


joamaki (approved):
Thanks!

PR #159, opened 2026-05-20
Do not panic on duplicate keys

jrajahalme:
Callers may create `index.KeySet`s that contain duplicate keys. Do not panic if this is the case.

The added test in the new `part_index_test.go` fails if applied without the change in `part_index.go`.


jrajahalme:
CI test passed, but looks like a PR from a fork has no access to setting PR comments?

eufriction:
Original Cilium issue for reference: https://github.com/cilium/cilium/issues/45758

joamaki (approved):

PR #162, opened 2026-06-10
[v0.4] part: Only reuse watches hash map when it is small

joamaki:
[ upstream commit d02e7291d2f352c099a281b3c3deb053560a4cd8 ]

This avoids carrying over a large allocation [txn.watches] from a massive one-off transaction into future transactions.

This should mitigate https://github.com/cilium/cilium/issues/44310.

### DIFF
--- a/part/txn.go
+++ b/part/txn.go
@@ -168,6 +168,10 @@ func (txn *Txn[T]) CommitOnly() *Tree[T] {
 	return t
 }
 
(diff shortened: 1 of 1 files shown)


joamaki:
Backporting this to v0.4 to also have the partial workaround for 44310 in Cilium v1.18 which still tracks StateDB v0.4.

pippolo84 (approved):
✔️

PR #163, opened 2026-06-11
statedb: avoid revision bumps for missing deletes

tklauser:
Currently, `(*writeTxnState).delete` increments the table revision before checking whether the target object even exists. A committed delete of a missing object could thus advance the table revision without producing a corresponding revision index entry or watch notification.

Fix this by moving the revision allocation until after the primary key lookup succeeded and any `CompareAndDelete` guard passed.


joamaki (approved):

joamaki:
Thanks! Just missed StateDB v0.8.2. Should've checked the open PRs first..

PR #164, opened 2026-06-11
script: introduce db/dump script command 

giorio94:
Introduce a `db/dump` script command that dumps the entire contents of StateDB as JSON. This command mimics the behavior of the already existing `/dump/` HTTP API, and is intended to be used to retrieve a snapshot of the DB (for instance during sysdump collection), without having to register the HTTP API, and implement a custom client.

I'd additionally nominate this PR for backport to the v0.5 series, which is still used by Cilium v1.19, given that it has no regression risks being only a script command, but it will help with possible troubleshooting, given that it is the first Cilium version in which the operator also includes some statedb usage.  

Related: cilium/cilium#46503

### DIFF
--- a/script.go
+++ b/script.go
@@ -38,6 +38,7 @@ func ScriptCommands(db *DB) hive.ScriptCmdsOut {
 		"db/lowerbound":  LowerBoundCmd(db),
 		"db/watch":       WatchCmd(db),
 		"db/initialized": InitializedCmd(db),
+		"db/dump":        DumpCmd(db),
 	})
 }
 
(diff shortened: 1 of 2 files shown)


joamaki (approved):
LGTM. 

Perhaps we could think about just removing the HTTP stuff completely once we've switched over to this? As far as I remember it wasn't used for anything else except the dumping. Would also allow removing the 'AnyTable/RemoteTable' thing. 

EDIT: Oh wait, 'cilium-dbg/cmd/status.go' uses the 'RemoteTable' construct. Maybe that could be rewritten to use a shell command instead (it's just doing basically the 'health' command as part of 'cilium status').

giorio94:
> Perhaps we could think about just removing the HTTP stuff completely once we've switched over to this?

Yeah, I was wondering about the same as well. It shouldn't be difficult to rewrite the current usages to directly invoke the corresponding shell command once this is merged and the dependency bumped on the Cilium side. The actual removal will likely need to be postponed to a new minor version, though, as it is technically a breaking change.

PR #166, opened 2026-06-24
statedb: panic with an error for immutable object violations

tklauser:
In `(*writeTxnState).modify`, panic with an error instead of a plain string so the recovered panic value implements the error interface while preserving the same diagnostic message. This makes the failure easier for recovery code to handle through the normal error path.

Also, use `require.PanicsWithError` in `TestDB_Insert_SamePointer` to assert the panic is the expected error value with the expected message. This catches unrelated panics and regressions back to string panic values.


pippolo84 (approved):
💯

PR #168, opened 2026-06-24
statedb: reject duplicate tables in write transactions

tklauser:
Passing the same table more than once to WriteTxn will currently deadlock when the transaction tries to acquire the same table lock repeatedly.

Avoid this by detecting duplicate table arguments before locking and panic'ing in that case, making the misuse immediately visible to the user.

### DIFF
--- a/db.go
+++ b/db.go
@@ -193,12 +193,24 @@ func (db *DB) WriteTxn(tables ...TableMeta) WriteTxn {
 	txn := db.writeTxnPool.Get().(*writeTxnState)
 	txn.db = db
 
+	// Deduplicate the set of tables to avoid acquiring the same table lock
+	// more than once, which would deadlock down the line.
+	seen := make(map[int]struct{}, len(tables))
+	tables = slices.DeleteFunc(slices.Clone(tables), func(table TableMeta) bool {
+		pos := table.tablePos()
+		if pos < 0 {
+			panic(tableError(table.Name(), ErrTableNotRegistered))
+		}
+		if _, exists := seen[pos]; exists {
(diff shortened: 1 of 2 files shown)


tklauser:
Another options would be to de-duplicate the tables in the write transaction instead of panicing. But it seems tables are usually hard-coded so the current implementation of the PR would force users to adjust their `WriteTxn` calls instead of silently accepting duplicates. But happy to change if reviewers think otherwise.

db.go
```diff
@@ -193,12 +193,22 @@ func (db *DB) WriteTxn(tables ...TableMeta) WriteTxn {
 	txn := db.writeTxnPool.Get().(*writeTxnState)
 	txn.db = db
 
+	// Check for duplicates in the set of tables to avoid deadlocks down the line.
+	seen := make(map[int]struct{}, len(tables))
+	for _, table := range tables {
+		pos := table.tablePos()
+		if pos < 0 {
+			panic(tableError(table.Name(), ErrTableNotRegistered))
```
derailed, line 202: I am usually not a fan of panicing especially in case the user mistakenly passed in a dup. Seems a bit heavy vs detecting if tables has a dup. So either we remove the dup or return an error and let the call site decide how to handle it??

tklauser: Yeah, agree. Changed to deduplicate the tables. I don't think we can change the public API to return an error because that would be too intrusive. So I kept the existing panic and just moved it up given we iterate the tables here. PTAL.

derailed (approved):
@tklauser Thank you for the update!

PR #169, opened 2026-06-26
statedb: reject indexes with empty names

tklauser:
Querying a table with an empty index name currently falls back to the primary index. Change the semantics so that `NewTableAny` validates that the index name is non-empty.

### DIFF
--- a/table.go
+++ b/table.go
@@ -116,6 +116,9 @@ func NewTableAny[Obj any](
 	indexPos := SecondaryIndexStartPos
 	for _, indexer := range secondaryIndexers {
 		name := indexer.indexName()
+		if name == "" {
+			return nil, tableError(tableName, ErrEmptySecondaryIndexName)
+		}
 		anyIndexer := toAnyIndexer(indexer, indexPos)
 		table.secondaryAnyIndexers = append(table.secondaryAnyIndexers, anyIndexer)
 		table.indexPositions[indexPos] = name
--- a/table.go
+++ b/table.go
@@ -202,6 +202,11 @@ func (t *genTable[Obj]) released() {
 }
 
 func (t *genTable[Obj]) indexPos(name string) int {
+	// An empty index name refers to the primary index, matching getIndexer.
+	if name == "" {
+		return PrimaryIndexPos
+	}
+
 	// By default don't consider the internal indexes.
 	start := PrimaryIndexPos
 
--- a/table.go
+++ b/table.go
@@ -80,6 +80,11 @@ func NewTableAny[Obj any](
 		return nil, err
 	}
 
+	// Primary index must always be unique
+	if !primaryIndexer.isUnique() {
+		return nil, tableError(tableName, ErrPrimaryIndexNotUnique)
+	}
+
 	toAnyIndexer := func(idx Indexer[Obj], pos int) anyIndexer {
 		return anyIndexer{
 			name:          idx.indexName(),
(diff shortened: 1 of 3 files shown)


table.go
```diff
@@ -199,6 +199,11 @@ func (t *genTable[Obj]) released() {
 }
 
 func (t *genTable[Obj]) indexPos(name string) int {
+	// An empty index name refers to the primary index, matching getIndexer.
+	if name == "" {
```
pippolo84, line 212: `indexPos` is executed for secondary indexes too. In case of a secondary index with an empty name we'll enter this `if` returning the primary index.
I think we should explicitly reject secondary indexes with empty names in `NewTableAny`.

tklauser: > I think we should explicitly reject secondary indexes with empty names in `NewTableAny`.

I think rejecting secondary indexes with empty names is a worthwhile change. Fixed as suggested.

But AFAICS it still won't fix the issue here in all cases because the name passed to `indexPos` by callers may be the one extracted from `Query.index` or `QueryRequest.Index` and that could still be empty in some cases, depending on how the query was constructed.

pippolo84: I see.

Besides, looking again at the code I see that stateDB explicitly supports objects with an empty primary key (see `TestDB_EmptyKeys`) and with this change it is possible to have something like this:

```go
emptyPrefixIndex := NetIPPrefixIndex[*testObject]{
	Name:   "empty-prefix",
	Unique: true,
	FromObject: func(*testObject) iter.Seq[netip.Prefix] {
		return func(func(netip.Prefix) bool) {}
	},
}

table, err := NewTable(
	db,
	"test",
	keyIndex,         // Primary index
	emptyPrefixIndex, // Secondary index
)
```

Now suppose that this object with an empty primary key is in the table:

```go
testObject{
	Key:  "",
	Tags: part.NewSet("test-object"),
}
```

Then this query:

```go
queryObject := &testObject{Key: "unrelated"}
q := emptyPrefixIndex.QueryFromObject(queryObject)

table.Get(rtxn, q)
```

returns the stored testObject, despite the object we used to generate the query had a Key value equal to "unrelated" (IOW: from a primary key POV it should not match).

Even if I don't think it is dangerous, it feels inconsistent. Getting this right seems hard, that's why I wonder if we should instead return a "no-match" representation instead of steering an empty index name toward the primary index.

tklauser: Thanks, that's a good point. I didn't consider this in my original change. I agree that getting this right is rather tricky, so I'll move the PR to draft for now and will discuss it with @joamaki after he's back.

joamaki: Hmm I don't see how that `emptyPrefixIndex.QueryFromObject` would end up using the primary index (https://github.com/cilium/statedb/blob/main/part_index.go#L75). But yes we should validate that secondary indexes must have names. Actually we should probably always require non-empty index names. I think however it's fine defaulting to primary index in `indexPos` to support things like empty index name in HTTP queries (https://github.com/cilium/statedb/pull/173). 

@tklauser would you mind changing this PR to add validation that empty index names are not allowed? I already merged 173 so the empty index is allowed in indexPos().

tklauser: @joamaki thanks, changed as suggested to validate that index names are non-empty. PTAL.

pippolo84 (approved):

tklauser:
Extended the unit test a bit to cover both primary and secondary indexes.

joamaki (approved):

PR #170, opened 2026-07-01
statedb: fix RemoteTable HTTP error handling

immanuwell:
`RemoteTable` had 2 small footguns.

What broke
- `Changes()` on a missing table could return an empty stream and no error. The 404 JSON body got treated like a keepalive, kinda sneaky.
- `Get()` and `LowerBound()` could leave `resp.Body` open if the caller stops after the first item.

What changed
- check non-200 responses before streaming and return the server error
- close response bodies when iteration ends, even if iteration stops early

How to repro
1. Run `Test_http_RemoteTable_Changes_missing_table_returns_error`.
2. Before this patch, `NewRemoteTable[*testObject](base, "missing").Changes(ctx)` yields an empty stream and `nil` error.
3. After this patch, it returns `HTTP 404 Not Found: Table "missing" not found`.

1. Run `Test_http_RemoteTable_Get_closes_body_when_stopped_early`.
2. Before this patch, stopping after the first `Get()` result leaves the response body open.
3. After this patch, the body is closed right away.

Checked with `go test ./...`.

### DIFF
--- a/http_client.go
+++ b/http_client.go
@@ -49,6 +49,7 @@ func (t *RemoteTable[Obj]) query(ctx context.Context, lowerBound bool, q Query[O
 	// Use a channel to return errors so we can use the same Iterator[Obj] interface as StateDB does.
 	errChanSend := make(chan error, 1)
 	errChan = errChanSend
+	seq = emptySeq2[Obj, Revision]()
 
 	key := base64.StdEncoding.EncodeToString(q.key)
 	queryReq := QueryRequest{
@@ -60,6 +61,7 @@ func (t *RemoteTable[Obj]) query(ctx context.Context, lowerBound bool, q Query[O
 	bs, err := json.Marshal(&queryReq)
 	if err != nil {
 		errChanSend <- err
+		close(errChanSend)
 		return
 	}
 
@@ -67,6 +69,7 @@ func (t *RemoteTable[Obj]) query(ctx context.Context, lowerBound bool, q Query[O
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url.String(), bytes.NewBuffer(bs))
 	if err != nil {
 		errChanSend <- err
+		close(errChanSend)
 		return
 	}
 	req.Header.Add("Content-Type", "application/json")
@@ -75,9 +78,15 @@ func (t *RemoteTable[Obj]) query(ctx context.Context, lowerBound bool, q Query[O
 	resp, err := t.client.Do(req)
 	if err != nil {
 		errChanSend <- err
+		close(errChanSend)
 		return
 	}
-	return remoteGetSeq[Obj](json.NewDecoder(resp.Body), errChanSend), errChan
+	if resp.StatusCode != http.StatusOK {
+		errChanSend <- decodeHTTPError(resp)
+		close(errChanSend)
+		return
+	}
+	return remoteGetSeq[Obj](resp.Body, json.NewDecoder(resp.Body), errChanSend), errChan
 }
 
 func (t *RemoteTable[Obj]) Get(ctx context.Context, q Query[Obj]) (iter.Seq2[Obj, Revision], <-chan error) {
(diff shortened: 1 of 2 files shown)


http_test.go
```diff
@@ -195,3 +197,76 @@ func Test_http_RemoteTable_Changes(t *testing.T) {
...
+
+func Test_http_RemoteTable_Changes_missing_table_returns_error(t *testing.T) {
+	_, _, ts := httpFixture(t)
+
+	base, err := url.Parse(ts.URL)
+	require.NoError(t, err, "ParseURL")
+
+	remoteTable := NewRemoteTable[*testObject](base, "missing")
+
+	changes, errs := remoteTable.Changes(context.Background())
```
pippolo84: Nit: prefer t.Context() instead of context.Background() (same in `Test_http_RemoteTable_Get_closes_body_when_stopped_early`)

http_client.go
```diff
@@ -143,11 +152,17 @@ func (t *RemoteTable[Obj]) Changes(ctx context.Context) (seq iter.Seq2[Change[Ob
 		close(errChanSend)
 		return
 	}
-	return remoteChangeSeq[Obj](json.NewDecoder(resp.Body), errChanSend), errChan
+	if resp.StatusCode != http.StatusOK {
+		errChanSend <- decodeHTTPError(resp)
+		close(errChanSend)
+		return emptySeq2[Change[Obj], Revision](), errChan
```
pippolo84: This is not introduced by this PR, but an empty seq should actually be returned in case of any errors, both here and for Get and LowerBound. I suggest to initialize `seq` with `emptySeq2` (here and in `query`).

http_client.go
```diff
@@ -170,3 +185,18 @@ func remoteChangeSeq[Obj any](dec *json.Decoder, errChan chan error) iter.Seq2[C
 		}
 	}
 }
+
+func emptySeq2[A, B any]() iter.Seq2[A, B] {
+	return func(func(A, B) bool) {}
+}
+
+func decodeHTTPError(resp *http.Response) error {
```
pippolo84, line 187: Since we are gonna decode here any possible HTTP response where `StatusCode != StatusOK`, we can simplify `remoteGetSeq` to care only about decoding error, changing it with something like this:

```go
	return func(yield func(Obj, Revision) bool) {
		defer body.Close()
		defer close(errChan)
		for {
			var resp responseObject[Obj]
			if err := dec.Decode(&resp); err != nil {
				if errors.Is(err, io.EOF) {
					return
				}
				errChan <- fmt.Errorf("decode error: %w", err)
				return
			}
			if !yield(resp.Obj, resp.Rev) {
				return
			}
		}
	}
```

This is consistent with what `remoteChangeSeq` is already doing.

http_client.go
```diff
@@ -95,27 +103,28 @@ type responseObject[Obj any] struct {
...
+		defer close(errChan)
 		for {
 			var resp responseObject[Obj]
 			err := dec.Decode(&resp)
 			errString := ""
 			if err != nil {
 				if errors.Is(err, io.EOF) {
-					close(errChan)
-					break
+					return
```
pippolo84, line 114: Nit: this is not an error, but I don't think this is useful either: changing `break` with `return` does not have any effect (same for the other occurrences here) because there is nothing after the `for` loop

pippolo84 (commented):
Thanks for the PR!

Overall LGTM, but I left some feedback inline to improve it even further.

http_test.go
immanuwell: done

http_client.go
immanuwell: done

http_client.go
immanuwell: done in the latest push

http_client.go
immanuwell: got it, done

immanuwell:
@pippolo84 addressed all your suggestions, PTAL

pippolo84 (approved):
LGTM, thanks! ✔️

pippolo84:
AFAICT the test failure is unrelated to the changes, it is actually a permission problem while trying to comment the PR with tests result.

PR #171, opened 2026-07-09
statedb: extend Derive to propagate initialization signal

giorio94:
Extend [statedb.Derive] to additionally register an initializer, and mark it as initialized once the upstream table is also initialized. In other words, let's correctly propagate the initialization signal, to prevent that the downstream table is immediately treated as initialized, e.g., potentially causing an associated reconciler to trigger pruning to early, against entries that should instead be preserved.


pippolo84 (approved):
💯

PR #173, opened 2026-07-14
fix: reject unknown indexes in HTTP queries

immanuwell:
Unknown index names make `/query` panic and close the connection. Kinda rough.

Return `400 Bad Request` with a JSON error instead. Primary and `ByRevision` queries still work.

Repro: send `GET /query` with `{"key":"","table":"test","index":"unknown"}`. Before, it panics and the client gets EOF. Now it returns `400`.

### DIFF
--- a/http_test.go
+++ b/http_test.go
@@ -4,8 +4,11 @@
 package statedb
 
 import (
+	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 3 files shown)


http_test.go
```diff
@@ -114,6 +114,22 @@ func Test_http_runQuery(t *testing.T) {
 	}
 }
 
+func Test_http_query_unknown_index_returns_bad_request(t *testing.T) {
```
derailed: For completeness we should probably set up a table test for this and test other use cases i.e empty, revisionIndex, happy path, ...

derailed (changes requested):
@immanuwell Nice catch!

http_test.go
immanuwell: got it, addressed table setup, normal primary path, revision index, and unknown-index error.
but I didn't add requested empty-index case because it currently exposes separate shared panic already covered by draft PR #169
should I add this as well?

derailed: @immanuwell  Thank you for the updates! 
I think we should test for empty and we can tune the test later so we don't loose track

derailed (commented):
@immanuwell See comment below.

http_test.go
immanuwell: done, added the empty-index test and the required fix

immanuwell:
@derailed addressed your suggestions, PTAL

derailed (approved):
@immanuwell Thank you for the updates. LGTM. Added Jussi/Tobias for their review.

immanuwell:
@tklauser @joamaki please take a look

joamaki (approved):

PR #175, opened 2026-08-18
fix: close HTTP change iterator on disconnect

immanuwell:
A disconnected `/changes/{table}` client leaves its deletion tracker registered until a later runtime cleanup. 
Deleted objects can stay in the graveyard after the stream is gone

The handler now closes its iterator on exit. 
The existing end to end test checks the tracker goes away, no API changes

Repro:
1. Apply only the test hunk
2. Run `go test ./ -run Test_http_RemoteTable_Changes -count=1`
3. It times out waiting for zero trackers. Apply the fix and it passes

Tested with `make test` and `make test-race`

### DIFF
--- a/http.go
+++ b/http.go
@@ -170,6 +170,7 @@ func (h dbHandler) changes(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+	defer changeIter.Close()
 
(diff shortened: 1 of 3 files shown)


joamaki:
There is a GC finalizer on change iterators so it is fine not to call `Close()` but of course it is better to call it. Thanks!

joamaki (approved):

PR #176, opened 2026-08-19
Fix reconciler retrying when multiple reconcilers race on an object

joamaki:
If CompareAndSwap failed then we checked whether the status was still
pending and had same ID which signalled that even though the object had
changed the data we reconciled with hadn't and thus could just update the
status. If the reconciliation was a failure we were supposed to queue it
for retry if we successfully committed the status. Unfortunately we ended
up looking at the error from CompareAndSwap and assumed the insert failed
even though it had been successfully resolved.

This impacted objects that had multiple reconcilers. If another reconciler
went first it committed its status and caused the CompareAndSwap to fail
and while we did resolve and insert the status we skipped queuing the object
for retry.


bimmlerd (approved):
fix LGTM, haven't reviewed test in detail but seems sensible

PR #177, opened 2026-08-19
WatchSet.Wait now waits even with single channel

joamaki:
The original semantics of returning immediately when the channel in a single channel set closes is at odds with how this is being used.

We want the settle time to be a form of rate-limiting to avoid computing too often and this should also apply to the single channel case.

Change the WatchSet.Wait to also wait with a single channel set.

Reported-by: Marco Iorio <marco.iorio@isovalent.com>


bimmlerd (approved):
Right, so this means you can rely on it always taking `settleTime` and hence use it for rate limiting. Makes sense to me.

PR #179, opened 2026-08-23
fix: stop Set.All iteration when requested

immanuwell:
`Set.All()` keeps yielding after a caller stops early. 
With 2 values, a normal `break` panics with `range function continued iteration after function for loop body returned false`

Repro:

```go
for range part.NewSet("first", "second").All() {
	break
}
```

Before this panics. 
After this fix exits cleanly

The iterator now returns when `yield` is false, with a regression test

Tested with `make test`, `make test-race`, `make build` and `make bench`


joamaki (commented):
Oopsie! Thanks for the fix.

joamaki (approved):

PR #181, opened 2026-08-24
Small performance improvements

joamaki:
- Skip unnecessary work in `retries.Clear` when queue is empty
- Avoid constructing iterator when dealing with singleton `part.Set`s
- Speed up `part.Set[string]` key construction with unsafe (same trick as `index.String`)

See commits for more details. 
AIL:3.

### DIFF
--- a/index/set.go
+++ b/index/set.go
@@ -10,12 +10,12 @@ func Set[T any](s part.Set[T]) KeySet {
 	toBytes := s.ToBytesFunc()
 	switch s.Len() {
 	case 0:
-		return NewKeySet()
+		return KeySet{}
 	case 1:
-		for v := range s.All() {
-			return NewKeySet(toBytes(v))
+		if v, ok := s.First(); ok {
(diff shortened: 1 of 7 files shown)


derailed (approved):
@joamaki Nice work! It would be good to understand before/after cpu/mem benchmarks. ie how does tracking an extra element on the set saves in cpu/mem. I get the iterator concerns but seems tracking a set of one feels a bit odd and how often is this the case??

joamaki:
> @joamaki Nice work! It would be good to understand before/after cpu/mem benchmarks. ie how does tracking an extra element on the set saves in cpu/mem. I get the iterator concerns but seems tracking a set of one feels a bit odd and how often is this the case??

It avoids the allocation of `part.Tree` which is a significant saving. We do the same trick in https://github.com/cilium/statedb/blob/main/part/map.go and also in https://github.com/cilium/cilium/blob/main/pkg/container/set/set.go#L18. Single element sets are also quite common.

PR #183, opened 2026-08-26
lpm: reject disjoint prefix queries

immanuwell:
A disjoint `Prefix` query could return an unrelated LPM subtree when traversal hit a compressed node

Repro:
1. Insert only `10.0.0.0/8`
2. Query `Prefix` with `192.168.0.0/16`
3. The `10/8` object comes back

The fix returns no iterator when matching diverges inside a node. 
Added trie and public `NetIPPrefixIndex` regression tests

Tests: `make`


joamaki:
Thanks for the PR. LGTM. FYI, found some other issues in the lpm implementation with Codex and I'll do a follow-up PRs on those, so maybe hold off on further fixes until I've posted the PRs?

PR #188, opened 2026-08-26
LPM fixes

joamaki:
- Preserve all objects sharing a key in non-unique LPM indexes instead of overwriting earlier entries.
- Correct lookups for keys terminating within compressed trie paths.
- Deduplicate objects indexed under multiple matching prefixes in non-unique LPM query results.
- Canonicalize IPv4-mapped IPv6 to IPv4 while keeping ordinary IPv6 lookups isolated.
- Reject LPM indexes as primary table indexes because the LPM indexer doesn't implement all required methods for primary index.
- Prevent prefix-length arithmetic from overflowing for maximum-length LPM keys.

See commits for detailed descriptions.

### DIFF
--- a/lpm/key.go
+++ b/lpm/key.go
@@ -48,9 +48,13 @@ func NetIPPrefix4ToIndexKey(prefix netip.Prefix) index.Key {
 	)
 }
 
+func lpmDataLen(prefixLen PrefixLen) int {
+	return (int(prefixLen) + 7) / 8
+}
+
 func EncodeLPMKey(data []byte, prefixLen PrefixLen) index.Key {
-	dataLen := (prefixLen + 7) / 8
-	if int(dataLen) > len(data) {
+	dataLen := lpmDataLen(prefixLen)
+	if dataLen > len(data) {
 		panic(fmt.Sprintf("invalid LPM key, data too short (%d) for prefix length (%d)", len(data), prefixLen))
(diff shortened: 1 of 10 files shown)


joamaki:
cc @immanuwell

lpm/key.go
```diff
@@ -48,9 +48,13 @@ func NetIPPrefix4ToIndexKey(prefix netip.Prefix) index.Key {
...
+func lpmDataLen(prefixLen PrefixLen) int {
+	return (int(prefixLen) + 7) / 8
+}
+
 func EncodeLPMKey(data []byte, prefixLen PrefixLen) index.Key {
-	dataLen := (prefixLen + 7) / 8
-	if int(dataLen) > len(data) {
+	dataLen := lpmDataLen(prefixLen)
+	if dataLen > len(data) {
 		panic(fmt.Sprintf("invalid LPM key, data too short (%d) for prefix length (%d)", len(data), prefixLen))
```
derailed: Not a fan of killing the application should this occurs. Would it be better off to let the call site decide the next course of action?

derailed (approved):
@joamaki Nice work!

lpm/key.go
joamaki: Yeah good point. Since this is exported function it makes sense.

PR #190, opened 2026-08-27
reconciler: Add StatusSet.IsPendingOrRefreshing and IsDone

joamaki:
Add methods to efficiently query the set.
Change All() to return an iterator instead of producing a map.


pippolo84 (approved):
LGTM

PR #192, opened 2026-08-29
fix: handle HTTP query write errors

immanuwell:
HTTP query can panic after a client stops reading a streamed response

Return the write error and stop the query instead

Repro:
1. Add the regression test
2. Run `go test ./ -run Test_http_query_write_error_does_not_panic -count=1`
3. On main it panics on a disconnected writer. This keeps it quiet

Tests: `make test`, `make test-race`


joamaki (commented):
Heh why did I use panic there :D. Thanks for the fix.

PR #193, opened 2026-09-01
fix: remove obsolete empty secondary index keys

immanuwell:
Deleting an object with an empty secondary key leaves it queryable through that index

`KeySet.Exists` now treats a zero `KeySet` as empty, consistent with `Foreach`

Repro:
1. Apply the regression test only
2. Run `go test ./ -run TestDB_DeleteEmptySecondaryKey -count=1`
3. On main the deleted object still comes back

Tests: `make`

### DIFF
--- a/db_test.go
+++ b/db_test.go
@@ -1386,6 +1386,25 @@ func TestDB_EmptyKeys(t *testing.T) {
 
 }
 
+func TestDB_DeleteEmptySecondaryKey(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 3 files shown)


joamaki (approved):
Good catch, thanks!

joamaki:
Ah fun, this made me stumble into another bug: `index.String("")` may return nil which then would get interpreted as not existing which of course isn't what we want. It might be we actually need to not check for nil but actually keep track of whether the "head" is set or not separately. Will do a PR.

PR #194, opened 2026-09-02
Fix indexing of nil keys

joamaki:
index.KeySet was a struct{head: Key, tail []Key} and used head == nil as indication that the set was empty. This broke indexing of for example index.String("") as that returned nil as the key.

Fix this by adding a 'hasHead' to the KeySet struct. Add Len() to KeySet and change First() to return (Key, bool) so that caller can distinquish whether a nil key was present or not. The API change to index.First is minor and I found no use of it in cilium/cilium so going to include this in a patch release.

AIL:3

### DIFF
--- a/part_index.go
+++ b/part_index.go
@@ -20,8 +20,8 @@ type Index[Obj any, Key any] struct {
 	Name string
 
 	// FromObject extracts key(s) from the object. The key set
-	// can contain 0, 1 or more keys. Must contain exactly one
-	// key for primary indices.
+	// can contain 0, 1 or more keys. Primary indices may contain
+	// zero or one key; an object with zero keys is not indexed.
 	FromObject func(obj Obj) index.KeySet
 
 	// FromKey converts the index key into a raw key.
@@ -73,7 +73,7 @@ func (i Index[Obj, Key]) Query(key Key) Query[Obj] {
 func (i Index[Obj, Key]) QueryFromObject(obj Obj) Query[Obj] {
 	return Query[Obj]{
 		index: i.Name,
-		key:   i.FromObject(obj).First(),
+		key:   i.objectToKey(obj),
 	}
 }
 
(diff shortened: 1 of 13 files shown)


part_index.go
```diff
@@ -88,7 +88,16 @@ func (i Index[Obj, Key]) QueryFromKey(key index.Key) Query[Obj] {
 }
 
 func (i Index[Obj, Key]) ObjectToKey(obj Obj) index.Key {
-	return i.FromObject(obj).First()
+	return i.objectToKey(obj)
+}
+
+func (i Index[Obj, Key]) objectToKey(obj Obj) index.Key {
+	keys := i.FromObject(obj)
+	if keys.Len() != 1 {
```
joamaki: Not correct. We need to allow primary indexer to also reject indexing of objects.

part_index.go
```diff
@@ -88,7 +88,16 @@ func (i Index[Obj, Key]) QueryFromKey(key index.Key) Query[Obj] {
...
-	return i.FromObject(obj).First()
+	return i.objectToKey(obj)
+}
+
+func (i Index[Obj, Key]) objectToKey(obj Obj) index.Key {
+	keys := i.FromObject(obj)
+	if keys.Len() > 1 {
+		panic("primary index must return at most one key")
+	}
+	key, _ := keys.First()
```
pippolo84: I see we are discarding the boolean that differentiate between no key and a "zero-length" key. Are there possible consequences?

write_txn.go
```diff
@@ -128,13 +128,16 @@ func (txn *writeTxnState) modify(meta TableMeta, guardRevision Revision, newData
...
-	revision := table.revision
 
 	// Update the primary index first
-	obj := object{data: newData, revision: revision}
+	obj := object{data: newData}
 	idIndexTxn := txn.mustIndexWriteTxn(meta, PrimaryIndexPos)
-	idKey := idIndexTxn.objectToKey(obj)
+	idKey, indexed := idIndexTxn.objectToKey(obj)
+	if !indexed {
+		return object{}, false, nil, nil
```
pippolo84, line 137: Not sure it really matters, but do we want to check against `guardRevision` here too in order to return `ErrRevisionNotEqual` instead of a `nil` error?

pippolo84 (commented):
Nice catch. Overall LGTM, left a couple questions inline

part_index.go
joamaki: Good catch! Yeah we should actually change this to return `(index.Key, bool)`. It is breaking API change in cilium/cilium but the fixes are one-liners so I think I'll just go and do the change.

write_txn.go
joamaki: If `objectToKey` says the object is not to be indexed then there's really nothing to do here. It's not really a `ErrRevisionNotEqual` error since there is no existing object to compare against. I think it's fine that this is just a no-op.

pippolo84 (approved):
💯

PR #196, opened 2026-09-03
reconciler: Fix StatusSet YAML marshalling

joamaki:
As the fields of StatusSet were unexported and we didn't have a MarshalYAML it was simply marshalled as "{}". Add a MarshalYAML and extend tests to verify and while at it extend the existing Status marshalling tests to also cover YAML.


pippolo84 (approved):
✔️

PR #198, opened 2026-09-07
part: allocate watches fully lazily

joamaki:
The watch state was still allocated eagerly. This refactores the code to fully allocate watches lazily. The nodes now contain a 'lazyWatchChannel' which is an atomic pointer to 'atomicWatchPointer', so essentially each node has 'atomic.Pointer[atomic.Pointer[*runtime.hchan]]'. The indirection is required as a logical watch may be shared when the tree forks. If we'd just store 'atomic.Pointer[*runtime.hchan]' we wouldn't be able to avoid a double-close when the tree forks as the two instances would not be able to synchronize.

```
goos: linux
goarch: arm64
pkg: github.com/cilium/statedb/part
              │  before   │              after              │
              │  sec/op   │  sec/op     vs base             │
_Insert          112.1µ      111.1µ            ~ (p=0.280 n=10)
_WatchReplace   108.39µ      88.58µ      -18.27% (p=0.001 n=10)
_GetWatch        17.91µ      19.75µ      +10.28% (p=0.001 n=10)

              │  before   │              after              │
              │   B/op    │   B/op      vs base             │
_Insert         82.28Ki     74.31Ki       -9.68% (p=0.000 n=10)
_WatchReplace   63.81Ki     55.96Ki      -12.30% (p=0.000 n=10)
_GetWatch         0.000       0.000            ~ (p=1.000 n=10)

              │  before   │              after              │
              │ allocs/op │ allocs/op   vs base             │
_Insert          3.067k      2.047k      -33.26% (p=0.000 n=10)
_WatchReplace    2.011k      1.006k      -49.98% (p=0.000 n=10)
_GetWatch         0.000       0.000            ~ (p=1.000 n=10)
```

Reconciler benchmark (best of 3) before:

```
1000000 objects reconciled in 1.38 seconds (batch size 1000) 
Throughput 727151.51 objects per second
568MB total allocated, 6015247 in-use objects, 239MB bytes in use
```

After:
```
1000000 objects reconciled in 1.34 seconds (batch size 1000) 
Throughput 748576.38 objects per second
552MB total allocated, 5010271 in-use objects, 231MB bytes in use
```

An isolated comparison of the pointer wrapper with the direct channel pointer shows first channel creation improving from 51.44ns to 42.41ns (-17.54%), 128B to 120B (-6.25%), and 3 to 2 allocations (-33.33%). Accessing an existing channel changes from 1.999ns to 2.042ns (+2.15%).

I also compared against an equivalent atomic.Value implementation. atomic.Value was 31.98% slower when creating the first channel, 6.48% slower when loading an existing channel, 58.14% slower when closing before observation, and 31.54% slower for channel creation followed by close. Allocation counts were equal, but atomic.Value used 128B instead of 120B for channel creation and 16B instead of 8B for close before observation. So I would argue the trade-off of using `unsafe.Pointer` to grab the `*hchan` is acceptable. Yes it means StateDB won't work on Go implementations/architectures where `chan T` is not a pointer, but considering the other uses of `unsafe.Pointer` in `part/node.go` this isn't really an issue. It would also mean that if Go ever changes the internal representation things would break here, but tests will easily catch that if it were to happen and we can then change to `atomic.Value`.

After this PR the `part.RootOnlyWatch` option doesn't help nearly as much anymore. However removing it completely did increase allocations from 552MB to 581MB in the reconciler benchmark without impact on throughput, so I'll leave it around for now. We can consider removing it in the future to simplify the implementation.

AIL:3


part/node.go
```diff
@@ -202,34 +202,56 @@ func (n *header[T]) clone(watch bool) *header[T] {
...
+	h.flags = n.flags
+	h.prefixLen = n.prefixLen
+	h.prefixP = n.prefixP
+	if !watch {
+		h.watch.disable()
+	}
+}
+
+func (n *header[T]) cloneWithSharedWatch() *header[T] {
+	clone := n.clone(false)
```
giorio94: Nit. If I'm not mistaken, passing true would avoid calling `h.watch.disable()`, which is overridden anyway immediately below.

giorio94 (approved):
Changes look reasonable to me, although I definitely don't have a very deep knowledge of all the statedb internals. These changes are also making the logic a notch more difficult to read and reason about, but I guess that's the tradeoff to try to extract some more performance out of it.

part/node.go
joamaki: ah yep, that'll save an atomic store.

PR #199, opened 2026-09-08
fix: handle unsupported HTTP change stream flushing

immanuwell:
The `/changes` handler assumes every response writer implements `http.Flusher`. 
Middleware wrappers may not preserve this optional interface, so the stream can panic after writing its initial changes

Use `http.ResponseController.Flush` and return cleanly if flushing fails. 
The fix keeps wrapped handlers from taking a spill

Reproduce on `main`:
```
go test ./ -run Test_http_changes_without_flusher_does_not_panic -count=1
```

The test panics without the fix and passes with it

### DIFF
--- a/http.go
+++ b/http.go
@@ -187,7 +187,9 @@ func (h dbHandler) changes(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 		}
-		w.(http.Flusher).Flush()
+		if err := http.NewResponseController(w).Flush(); err != nil {
+			return
+		}
 		select {
 		case <-r.Context().Done():
 			return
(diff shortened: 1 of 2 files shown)


joamaki (approved):
Looks good to me. BTW, for future submissions it'd be great if you'd:

- read the AI policy at https://github.com/cilium/community/blob/main/AI-POLICY.md. If you're using AI to contribute include the AI influence level in your commit description and PR description (just e.g. "AIL:3" suffices). Write the PR description yourself. 
-  the commit message should be more informative than the PR description. Your commit messages have been empty thus far.

PR #201, opened 2026-09-11
db: introduce support for commit hooks

giorio94:
Introduce a new [statedb.New] option that allows to register commit hooks that get invoked every time that a transaction is committed, before releasing the associated locks. The hooks are strictly read only, and cannot abort the transaction. Additionally provide the possibility of registering the hooks through [statedb.Cell].

One possible use-case for the hooks is to validate properties of the entries of specific tables, for testing purposes. However, they may also unlock other use-cases, including the possibility of synchronously dumping the state to disk, and so on.

### DIFF
--- a/db.go
+++ b/db.go
@@ -97,6 +97,7 @@ type dbState struct {
 	gcRateLimitInterval time.Duration
 	metrics             Metrics
 	writeTxnPool        sync.Pool
+	commitHooks         []CommitHook
 }
 
 type dbRoot = []*tableEntry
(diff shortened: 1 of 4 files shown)


db.go
```diff
@@ -113,6 +115,24 @@ func WithMetrics(m Metrics) Option {
 	}
 }
 
+// CommitHook is a commit hook that can be registered through [WithCommitHooks].
+// Commit hooks are run synchronously, must be strictly read-only, and fast.
+// The given list of tables is only valid until the hook terminates, and must
+// explicitly copied if longer retention is needed.
+type CommitHook func(txn ReadTxn, tables ...string)
```
joamaki: let's just pass `tables []string` and not make this vararg.

joamaki (approved):

db.go
giorio94: Sure, updated.

PR #202, opened 2026-09-13
fix: handle indexes without string conversion

immanuwell:
Indexes can omit `FromString`, but `db/get` calls it anyway and panics

Repro:

1. Register an index without `FromString`
2. Run `db/get test 1`
3. Current main panics in `Index.fromString`

The fix: return a clear query error instead

Tests: `make`


joamaki:
See my comment at https://github.com/cilium/statedb/pull/199#pullrequestreview-5178800380. The commit description and the AI influence level disclaimer is missing. I'll add a PR template soon to have a checklist...

part_index.go
```diff
@@ -49,6 +50,9 @@ func (i Index[Obj, Key]) isUnique() bool {
 
 // fromString implements Indexer.
 func (i Index[Obj, Key]) fromString(s string) (index.Key, error) {
+	if i.FromString == nil {
+		return nil, fmt.Errorf("index %q does not support string queries", i.Name)
```
tommyp1ckles, line 54: @immanuwell Aside from @joamaki comments regarding AI usage this seems to make some sense, can we describe why this is needed? Did the table panic previously without this guard ?

immanuwell: @tommyp1ckles index without FromString registers successfully, but db/get then reaches the nil conversion and panics. and this guard returns a query error instead

immanuwell:
@joamaki I am sorry, added both commit description and AIL

immanuwell:
@tommyp1ckles everything is addressed, please take a look

joamaki (approved):

